PR #4, opened 2017-03-03
Use default-directory variable instead of buffer-file-name to pick root

sw9:
Currently, treemacs-init doesn't initialize the root correctly in dired-mode buffers because (buffer-file-name (current-buffer)) is always nil for such buffers. Looking at the buffer's default-directory is a more reliable way of getting the root. Also, it avoids the need to check if the buffer-file-name is a file or a directory.


Alexander-Miller:
:+1:

PR #5, opened 2017-03-04
do not remove slash if path only contains slash

sw9:
Handle case when treemacs is initialized in root directory


Alexander-Miller:
Damn slashes :/

PR #6, opened 2017-03-07
Two changes relating to deletion

sw9:
Fixed a bug with treemacs--clear-from-cache. Basically, it's possible when you delete a folder that it isn't   in the cache at all but this still calls the clear-from-cache function. And (= 1 nil) results in an error.

Also, added prompts to kill opened buffers after deleting a file/folder.

### DIFF
--- a/treemacs.el
+++ b/treemacs.el
@@ -400,9 +400,10 @@ Also remove any dirs below if PURGE is given."
   (let* ((parent (treemacs--parent path))
          (cache  (assoc parent treemacs--open-dirs-cache))
          (values (cdr cache)))
-    (if (= 1 (seq-length values))
-        (setq treemacs--open-dirs-cache (delete cache treemacs--open-dirs-cache))
-      (setcdr cache (delete path values)))
+    (when values
+      (if (= 1 (seq-length values))
+          (setq treemacs--open-dirs-cache (delete cache treemacs--open-dirs-cache))
+        (setcdr cache (delete path values))))
     (when purge
       (-if-let (children (-flatten (--map (cdr (assoc it treemacs--open-dirs-cache)) values)))
           (progn
--- a/treemacs.el
+++ b/treemacs.el
@@ -855,6 +855,7 @@ Do nothing for directories."
                 (f-delete path t)
                 (treemacs--clear-from-cache path t)
                 t))))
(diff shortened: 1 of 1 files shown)


Alexander-Miller:
You need to also require dired - the cleanup isn't autoloaded.

sw9:
hmm that's true. do you think it's better to require dired and adding that dependency or just inline the necessary portion directly? It's a pretty short snippet.

Alexander-Miller:
Is the dired cleanup even what's needed here? As I see it what you're going for is asking to also kill every buffer that, after the deletion, is associated with a deleted file.

However in the following example, having opened file1/2/3 and then letting treemacs delete foo I get nothing. No query and the opened buffers stick around.

http://i.imgur.com/NqrQGBc.png

Maybe we're better off doing our own thing? Just go through the entire buffer-list, and aks to kill every buffer where path is a prefix of buffer-file-name. Plop it down in misc. utils and we're good to go.

At any rate you'll need to grab the recent commits. Your touching up on delete made to do some refactoring of my own.

sw9:
ah looks like the reason it doesn't do anything for you is that dired cleanup makes you load dired-x as well, otherwise, it doesn't work even though it doesn't really depend on anything from that package. Presumably you don't have this loaded, whereas I've already loaded it earlier in my init file for another purpose. In this case, probably better to do our down thing then.

Alexander-Miller:
Nope, it's loaded, `(and (featurep 'dired-x) dired-clean-up-buffers-too)` returns t. Looking at the code of dired-cleanup it just doesn't do what we want. What it does is kill the buffer visiting `path` and then kills all *dired mode* buffers for `path`, but not buffers visiting files in `path`.

Doing our own thing it is then.

sw9:
Alright sounds good, though the first change is still needed. I just fetched the latest from your repo, and tried to delete a folder, and I still get the error.

Alexander-Miller:
FYI: I've pushed the open-dir-cache fix now. It's also part of the test suite.

sw9:
Ah, I was just going to open a new MR when I had it done, but leaving this open to discuss is fine too. Any preference for the name? Was thinking treemacs-kill-buffers-after-deletion.

Alexander-Miller:
Needs a 'private' double dash, otherwise it's good.

PR #7, opened 2017-03-16
treemacs--kill-buffers-after-deletion

sw9:


treemacs.el
```diff
@@ -605,6 +605,58 @@ Delete all elements whose car is ‘eq’ to KEY from ALIST."
...
+         (y-or-n-p (format "Kill buffer %s in %s, too? "
+                           (buffer-name)
+                           (file-name-nondirectory path)))
+         (kill-buffer buf)
+         )
+        )
+      )
+
+    ;; Kill all dired buffers in one step
+    (let (dired-buffers-for-path)
```
Alexander-Miller: Not quite a single thread, but I'd much prefer to change up the code to use the same dash.el style as elsewhere:

    (when-let (dired-buffer-for-path 
               (->> dired-buffers
                    (--filter (treemacs--path-in-dir (car it)))
                    (-map #'cdr)))
      (and (y-or-n-p (format "Kill Dired buffers of %s, too? " (f-filename path)))
           (-each dired-buffers-for-path #'kill-buffer)))

treemacs.el
```diff
@@ -605,6 +605,58 @@ Delete all elements whose car is ‘eq’ to KEY from ALIST."
...
+  (if (not is-dir)
+      (let ((buf (get-file-buffer path)))
+        (and buf
+             (y-or-n-p (format "Kill buffer of %s, too? "
+                               (file-name-nondirectory path)))
+             (kill-buffer buf)))
+
+    ;; Prompt for each buffer visiting a file in directory
+    (dolist (buf (buffer-list))
+      (with-current-buffer buf
```
sw9: I call buffer-file-name and buffer-name without specifying the buffer, so with-current-buffer is actually needed here. However, I'll just call those on the actual buffer then.

treemacs.el
sw9: Never used this macro before. Reading up on it.

treemacs.el
Alexander-Miller: Yes, either way works. With two aequivalent solutions I'd just go with the one introducing fewer new code blocks and indentation.

treemacs.el
Alexander-Miller: It's very useful if your chaining a bunch of method calls. Instead of `(foo (bar (baz (qux x))))` (which gets much worse when there's more than a single argument involed) you'd instead write `(-> x (qux) (baz) (bar) (foo))`. You'll find plenty of practical examples elsewhere in the code.

Other than that dash.el is also chock-full of other functional goodies like those -map, -each and -filter functions I use everywhere.

treemacs.el
```diff
@@ -605,6 +605,40 @@ Delete all elements whose car is ‘eq’ to KEY from ALIST."
   (beginning-of-line)
   (forward-char))
 
+(defun treemacs--path-in-dir? (path dir)
+  "Is PATH in directory DIR?"
+  (s-starts-with? (concat dir "/") path))
```
Alexander-Miller: Is adding the slash needed? I can't think of a case where it would make a difference. 
Being this short this function is also a good candidate for being a defsubst instead of a defun.

Everything else looks good. Just squash it up and we're good to go. Since we're adding a single feature IMO the changes should into one commit, or at most two as was the case originally.

sw9: I think it is. Suppose you have a folder "/folder/" which you delete and you also have some buffers open from "/folder-bah/". Not adding the slash would also close all the buffers from that folder as well, since dir names don't contain the last slash by default.

Alexander-Miller: True. Good catch!
Now only a defsubst and squash and we're go.

sw9:
done

PR #14, opened 2017-06-04
Fix broken package dependency version in treemacs-evil

purcell:
See https://github.com/melpa/melpa/pull/4760


Alexander-Miller:
Thanks 👍

PR #37, opened 2017-08-15
add js and json icon to treemacs

jbharat:
* icons are 22x22
* should be visible on both light and dark themes


Alexander-Miller:
Icons look great. MR needs a couple of changes though:

 * MR must be against master. The dev branch is a playground, filled with I don't know what.
 * I just pushed a new version, so you need to rebase. The new `treemacs-visit-node-default-action` now bound to RET should also be relevant to your interests.
 * The icons are a bit too large, in that the colored part pretty much fills out the entire icon area. This leaves little to none space between these 2 icons, which looks out of place compared to all the other icons used in treemacs. [You can use these icons instead](https://imgur.com/a/hvPSh). They're the same as yours, I just sized them down to 20x20 and then used the extra 2px as extra space.

jbharat:
- I have rebased with master
- I had access to photoshop so created new icons again as per the specification. Have a look at them. Let me know if you need any changes. The canvas size 22x22 and the image within is 20x20

Thanks for the RET. I was just about to get used to `oo`. Not to late.

Alexander-Miller:
Looks good. Squash the commits into 1 and I'll do the merge.

> Thanks for the RET. I was just about to get used to oo. Not to late.

I'm a backend dev. What's a UX?

Seriously though, oo is the most common case, but don't forget about ox or oav.

jbharat:
I think I messed up. I am not sure. I have never done squash. Followed this [link](https://stackoverflow.com/questions/2563632/how-can-i-merge-two-commits-into-one) and merged my two commits. I dont see any change though.

UX = user experience

I find myself using RET and oaa the most. I always have splits and oaa is a nice feature to set on a specific window.

Alexander-Miller:
> UX = user experience

I know, I was making a joke there. 

>  Followed this link and merged my two commits. I dont see any change though.

Something weird happened here - github says this PR is supposed to merge 15 different commits, most of them mine and already in master.

Your changes are simple. so we best try again with a clean slate: Delete your local branch and create it again based on my master, add in the new icons and their setup with a single commit and push --force to your master. That should clean things up here.

> I have never done squash. Followed this link and merged my two commits.

When youre squashing the last few commits on your branch IMO it's easiest to just get rid of them: e.g. `git reset --mixed HEAD^^^` will remove the last 3 commits, but keep the actual changes you made, so you can just stage all and make a single commit.

If it's anything more complicated than that I highly recommend Magit which has a nice interface for interactive rebases.

PR #46, opened 2017-09-15
Fix selected-icon-background to work with default Emacs theme

sw9:
I had a problem with the icon of the selected file disappearing.

Before:
screen shot 2017-09-15 at 3 40 09 pm

After:
screen shot 2017-09-15 at 3 32 44 pm

### DIFF
--- a/treemacs-visuals.el
+++ b/treemacs-visuals.el
@@ -48,11 +48,11 @@ after a theme change.")
   "The last button treemacs has highlighted.")
 
 (defvar treemacs--selected-icon-background
-  (face-attribute 'hl-line :background)
+  (face-attribute 'hl-line :background nil t)
   "Background for selected icons.")
 
 (defvar treemacs--not-selected-icon-background
-  (face-attribute 'default :background)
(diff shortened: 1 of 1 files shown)


Alexander-Miller:
I assume you're running emacs as a daemon? I ran into this problem just today when playing around with emacs & systemd.

The cause seems to be that treemacs is loaded while emacs is a daemon and no frame exists, so those color vars are not defined in a way the current code can grab, which somehow causes the selected icon to disappear. Not sure where those definitions come from when you go with inherit, but it's a good call anyway, and saves the need fo potentially more awkward workarounds.

The deleted lines 85-86 need to stay however. Those colors must be updated whenever a new theme is loaded.

sw9:
hl-line inherits from highlight, but at least for me, the background for hl-line inherits from highlight instead of being defined directly for some reason. Thus, with the previous code, I had unspecified for the background color. I'm don't think Emacs supports daemon mode on OSX actually, though I do use server-start.

Ah, I thought 85-86 were part of the adwaita workaround which this change would make obsolete, my bad.

PR #49, opened 2017-09-25
Added more C++ extensions

netromdk:
The very common ".cc" wasn't present so I've added the most common C++ extensions to the list.

### DIFF
--- a/treemacs-visuals.el
+++ b/treemacs-visuals.el
@@ -139,7 +139,7 @@ Insert VAR into icon-cache for each of the given file EXTENSIONS."
   (treemacs--setup-icon treemacs-icon-shell      "shell.png"      "sh" "zsh" "fish")
   (treemacs--setup-icon treemacs-icon-pdf        "pdf.png"        "pdf")
   (treemacs--setup-icon treemacs-icon-c          "c.png"          "c" "h")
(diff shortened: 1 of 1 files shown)


netromdk:
Treemacs is really great btw!

Alexander-Miller:
Thanks for the MR, it's a good reminder to finally look for a c++ icon that's not quite so hideous.

The C and H extensions are not going to work - extensions are always lowercased before their icon is fetched, and since treemacs allows matching on "extensions" like .gitignore and .Xdefaults putting in all combinations isn't much of an option.
Shouldn't they use the c icon anyway? Or do c++ compilers differentiate between .c and .C files?

netromdk:
It's mostly a historical thing with .C and .H which was important for case sensitive systems of the older generations. But I understand the to lower check so I'll just remove those two.

netromdk:
In regards to the icon, maybe [this](https://pluralsight.imgix.net/paths/path-icons/c-plus-plus-93c7ddd5cc.png) could be used?

Alexander-Miller:
Squash those 2 commits and we're good to fo.

The icon [did not take well to being scaled down](https://imgur.com/a/gKVWG) (I just used imagemagick), even if I were to clean up the borders, the pluses are barely visible.

I'll try adapting the icon from all-the-icons. It's what i did for the C icon. Its pluses are also barely visible, but other than that that it'll look nice and clean.

netromdk:
I'll just create a new one from a feature branch instead to be more clean. Sorry.

PR #73, opened 2017-10-09
Update next/prev line, optional count argument

duianto:


treemacs-interface.el
```diff
@@ -43,16 +43,18 @@
...
 
-(defun treemacs-next-line ()
-  "Goto next line."
-  (interactive)
-  (forward-line 1)
+(defun treemacs-next-line (&optional count)
+  "Goto next line.
+A COUNT argument, moves COUNT lines down."
+  (interactive "p")
+  (forward-line (or count 1))
```
Alexander-Miller: The `or` isnt even needed. count is set to 1, I assume on account of the interactive flag.

duianto: Your right, I did not know that. It's fixed.

PR #86, opened 2017-10-28
Fix incorrect icon background colors after enabling/disabling theme

sw9:
It's better to defadvice around enable/disable theme rather than load theme. After loading a theme, a user may enable or disable it, and treemacs won't match the new colors. On the other hand, load-theme calls enable theme, so it's not necessary to defadvice load-theme.


Alexander-Miller:
👍

PR #94, opened 2017-11-05
Fixed minor spelling type

zumbalogy:


Alexander-Miller:
There's a typo in the commit message and this is an irony free zone. Please fix.

Better yet, just squash the commits. There's no need to keep them separate.

zumbalogy:
Muphry's law strikes again https://en.wikipedia.org/wiki/Muphry%27s_law

Alexander-Miller:
👍

PR #95, opened 2017-11-13
Fix readme.org typos

duianto:
I saw a typo, so I did a quick flyspell check of the whole file.

---

I'll mention the typo on the wiki page here, since wiki pages don't seem to support pull requests.

On the Wiki page: https://github.com/Alexander-Miller/treemacs/wiki
The last word in the first bullet point is missing an `i`:
>If you want to get a feel for the codebase look to the functions treemacs and treemacs--push-button, they're general entry points for a large part of treemacs' functionalty.

before:
`functionalty`

after:
`functionality`

### DIFF
--- a/README.org
+++ b/README.org
@@ -66,7 +66,7 @@ If you don't care about reading the full readme here's a list of some bare bones
    [[file:hydra.png]]
  * If you use evil don't forget to also install ~treemacs-evil~
  * If you use projectile you can install ~treemacs-projectile~ to allow treemacs to interact with projects.
- * Treemacs doesn' bind any global keys, you need to use whatever fits you best. A full install setup for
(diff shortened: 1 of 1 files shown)


Alexander-Miller:
👍 

> I'll mention the typo on the wiki page here, since wiki pages don't seem to support pull requests.

Aren't wikis freely editable? Or is there some setting I need to change?

duianto:
>By default, only collaborators on your repository can make changes to wikis, but you can configure this to be enabled for all users on public repositories.

I found the quote in this answer: https://stackoverflow.com/questions/37776247/how-can-i-update-github-wiki-pages/37776291#37776291
it comes from this page: https://help.github.com/articles/about-github-wikis/

Alexander-Miller:
Ah, I looked through the settings before, but missed this part. The wiki should be all open now.

PR #99, opened 2017-11-15
Fix undefined variable error in popwin advice

mkcms:
Commit 3984a01c8b54509b6 didn't replace usage of removed variable
`treemacs--buffer-name`in advice to function
`popwin:{create,close}-popup-window` with
`treemacs--get-framelocal-buffer`.

### DIFF
--- a/treemacs-compatibility.el
+++ b/treemacs-compatibility.el
@@ -42,8 +42,8 @@
 
   (defadvice popwin:create-popup-window
       (around treemacs--popwin-popup-buffer activate)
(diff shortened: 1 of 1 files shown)


Alexander-Miller:
Thanks. I was going to fix this when I get back from work, but a PR is fine, too.

The best solution is, however, to just delete this block. It's needed to get around `window-size-fixed` being t, but treemacs does not use that option anymore.

mkcms:
@Alexander-Miller The popwin block is deleted now.

Alexander-Miller:
👍

PR #107, opened 2017-11-19
Added internal support for dragging a node to display in any window

rswgnu:
My Hyperbole package includes support for dragging list items (such as in dired or buffer menu) to a window in any frame and displaying the item there.  This simple pull request adds the internal capabilities to treemacs for the same capability, including for functions and variables within files.  I hope you will include it.

Then the next release of Hyperbole will add the few lines at the bottom of this variable setting to support treemacs-mode.
```
(defvar hmouse-drag-item-mode-forms
  (hmouse-map-modes-to-form
  '((Buffer-menu-mode (Buffer-menu-buffer t))
    (ibuffer-mode (ibuffer-current-buffer t))
    (helm-major-mode (helm-get-selection (current-buffer)))
    ;; Note how multiple major modes may be grouped with a single form for item getting.
    ((dired-mode vc-dired-mode wdired-mode) (or (when (dired-get-filename nil t)
						  (hmouse-dired-display-here-mode 1)
						  (dired-get-filename nil t))
						;; Drag from first line current directory
						;; means move this dired buffer to the
						;; release window.
						(prog1 (current-buffer)
						  (hmouse-pulse-buffer)
						  (bury-buffer))))
     (treemacs-mode (if (fboundp 'treemacs-node-buffer-and-position)
			(treemacs-node-buffer-and-position)
		      (error "(hmouse-item-to-window): Update the treemacs package for item dragging support")))))
  "List of (major-mode lisp-form) lists.
The car of an item must be a major-mode symbol.  The cadr of an item
is a Lisp form to evaluate to get the item name at point (typically a
buffer, file or directory name whose contents will be displayed in the
drag release window.")
```


Alexander-Miller:
Very interesting stuff, but the general way it's implemented does not fit into treemacs' conventions:

 1) It's glue code for a different package mixed with proper implementation code, and I try to keep them separate.
 2) The compiler is complaining about functions that aren't known to be defined, while I strive for a zero warnings code base.

These are minor issues and quickly fixed by moving the code to an after-eval block in treemacs-compatibility.el and peppering it with a few declare-function statements. However I am thinking that since this is explicitly glue, and not feature or fix code, used only for hyperbole, it would be better to extract the new functions into their own package. 
treemacs-evil and treemacs-projectile already follow this pattern and it would make it easier to write code that the compiler likes. Loading all of treemacs is much easier than getting through my admittedly convoluted web of inter-file dependencies. 
Seeing treemacs-hyperbole in the package list when looking for treemacs is also a bit more advertisement friendly.

What do you think?

rswgnu:
> However I am thinking that since this is explicitly glue, and not feature or fix code, used only for hyperbole, it would be better to extract the new functions into their own package.

I neglected to mention it but I think all the requested additions/changes could easily be used by another package that wants to control where Treemacs displays things, so I don't think its use would be limited in the end to Hyperbole.  Hyperbole will just be the first implementation using it.

If you don't want to put the new functions next to the functions from which they derive, then I could see just adding a new file to Treemacs but I think this is useful enough to go in the core distribution.  Of course it is up to you.  I am happy to fix any byte-compilation warnings that I can if you point them out.

> treemacs-evil and treemacs-projectile already follow this pattern and it would make it easier to write code that the compiler likes. Loading all of treemacs is much easier than getting through my admittedly convoluted web of inter-file dependencies.

If these changes provided the full drag-and-drop functionality, then I could see going that way but as is, they are just API-type functions for use by other packages.

> Seeing treemacs-hyperbole in the package list when looking for treemacs is also a bit more advertisement friendly.

True but I want people who install Treemacs and Hyperbole to get the benefit out-of-the-box.
I have already added further Treemacs support to Hyperbole for the next release (single mouse key to run most major functions in the Treemacs buffer) and added a file for Treemacs in the Hyperbole distribution.

If you decide you don't want to add this to Treemacs, then I'll just put those functions in the Hyperbole treemacs file but then only Hyperbole would benefit and changes you make in the future might break the previous code and we wouldn't know to update it until someone reported it or we noticed it in internal testing.

Happy to talk further about this if you would like and thanks for looking through the ideas I submit.

Alexander-Miller:
The changes being general mouse interface api is an argument I can agree with. Other than the simple click api implemented right now I haven't looked far into emacs' mouse interface, though I plan to get to it eventually - you've already contributed to the relevant discussion issue. Being able to drag a node to a buffer sounds interesting. Not very emacs-y, but very interesting.

The mouse interface code would have gone into its own treemacs-mouse-interface.el file. I'll push an empty one later tomorrow so you can just shift your changes there.

That just leaves one last issue: `(require 'hmouse-tag)`. That's a hard, undeclared dependency and must be dealt with properly. Making treemacs depend on hyberbole for it is overkill. Making it implicitly load after hyperbole looks like a no-go, since it's needed for mouse interaction with tags. 
The xref utils look small and simple and are only needed in 1 function, so I suggest defining them there in a cl-flet block.

rswgnu:
On Wed, Nov 22, 2017 at 3:45 PM, Alexander-Miller <notifications@github.com>
wrote:

> The changes being general mouse interface api is an argument I can agree
> with.
>
​Great.​

> Other than the simple click api implemented right now I haven't looked far
> into emacs' mouse interface, though I plan to get to it eventually - you've
> already contributed to the relevant discussion issue.
>
​Good.  I have found an issue with your use of mouse-1 where I can't get
the Treemacs window to stay selected when I click it on certain items
(likely tags)​

​, so visiting this topic would be good.
I would suggest you leave mouse-1 untouched and do Treemacs-specific
selection on mouse-2.​

> Being able to drag a node to a buffer sounds interesting. Not very
> emacs-y, but very interesting
> ​​
> .
> ​​
>
> ​You can emulate drags from the keyboard as well.

> The mouse interface code would have gone into its own
> treemacs-mouse-interface.el file. I'll pu
> ​​
> sh an empty one later tomorrow so you can just shift your changes there.
>
> ​Ok.​


> That just leaves one last issue: (require 'hmouse-tag). That's a hard,
> undeclared dependency and must be dealt with properly. Making treemacs
> depend on hyberbole for it is overkill. Making it implicitly load after
> hyperbole looks like a no-go, since it's needed for mouse interaction with
> tags.
> The xref utils look small and simple and are only needed in 1 function, so
> I suggest defining them there in a cl-flet block.
>
​Good idea.  I tried to get these added to xref but the package author
preferred having everyone use more complicated lower-level primitives
because he said the API was not yet finalized.
​

rswgnu:
> That just leaves one last issue: (require 'hmouse-tag).  The xref utils look small and simple
> and are only needed in 1 function, so I suggest defining them there in a cl-flet block.

I have implemented this and pushed the changes on this branch, so you should be good to go.

Alexander-Miller:
Mouse-interface file is up now.

rswgnu:
Hi Alexander:

I have looked at this pull request of mine and it is not clear to me which if any of the changed code you would like placed in treemacs-mouse-interface.el.  Maybe if you move any of your code that you think belongs there first, then I could see what you want there.  This pull request does not implement any mouse bindings; that part was implemented in Hyperbole as is done for many other modes.

In Hyperbole, there are just two Hyperbole-specific functions that bind mouse keys (Hyperbole uses global mouse keys, typically shift-mouse2 and shift-mouse-3 but can use mouse-2 and mouse-3).  When Hyperbole has mouse support for a major mode, it will override the mode-specific bindings on those keys.  If you want to make the Treemacs default mouse bindings as close as possible to those of Hyperbole, I can email you those 2 functions or you can just look at the hui-treemacs.el file in the latest Hyperbole release.

As soon as we get the code in this pull request merged, I'll remove these changed functions from that Hyperbole file in the next release (leaving just the global mouse bindings).

Alexander-Miller:
> I have looked at this pull request of mine and it is not clear to me which if any of the changed code you would like placed in treemacs-mouse-interface.el.

Everything.  As far as I can see all the changes you introduce serve to provide infrastructure for hyperbole's and eventually treemacs' own mouse interface, so as far as I'm concerned they all belong into mouse-interface.el.
Except the change to the execute-button-action macro, it shouldn't interfere with any existing code and can stay as is.
From looking around `treemacs-node-buffer-and-position` does not seem to be a user facing function, so it should be renamed to use a `treemacs--` prefix.

Looking at hyperbole's code I also see you used `eval-and-compile` to load treemacs. I really struggle to understand the meaning of this macro. Would you mind explaining its use-case?

rswgnu:
> Everything. As far as I can see all the changes you introduce serve to provide infrastructure for hyperbole's and eventually treemacs' own mouse interface, so as far as I'm concerned they all belong into mouse-interface.el.

That is fine with me as it will make maintaining any changes even easier.

> Except the change to the execute-button-action macro, it shouldn't interfere with any existing code and can stay as is.

Ok.

> From looking around treemacs-node-buffer-and-position does not seem to be a user facing function, so it should be renamed to use a treemacs-- prefix.

Internal function names are for use only by the package itself, not outside of it.  That function is critical to Hyperbole's or other packages use of Treemacs when displaying tags for example, so it should be left as a public function.  A user-facing function in Emacs is called a command and has an (interactive) form within it.  So the two concepts are separate.  That is my understanding, anyway.

> Looking at hyperbole's code I also see you used eval-and-compile to load treemacs. I really struggle to understand the meaning of this macro. Would you mind explaining its use-case?

Here's what the Emacs Lisp manual says.  Mainly, the idea is to get Emacs macros right at compile-time:

 Most uses of ‘eval-and-compile’ are fairly sophisticated.

     If a macro has a helper function to build its result, and that
     macro is used both locally and outside the package, then
     ‘eval-and-compile’ should be used to get the helper both when
     compiling and then later when running.

     If functions are defined programmatically (with ‘fset’ say), then
     ‘eval-and-compile’ can be used to have that done at compile-time as
     well as run-time, so calls to those functions are checked (and
     warnings about “not known to be defined” suppressed).

rswgnu:
I wrote:
> Internal function names are for use only by the package itself, not outside of it.

Then I looked at the Hyperbole mouse interface to Treemacs and found it had to call several functions that you have named as internal functions.  Maybe we could externalize these as these are very useful for interfacing to Treemacs.

They are:

treemacs--current-button
treemacs--is-treemacs-window?
treemacs--buffer-exists?

rswgnu:
Ok, the treemacs-mouse-interface pull request is now ready for you to test that everything remains fine with normal treemacs usage.  I did change the names of the other functions that Hyperbole needed to be public, as well.

rswgnu:
Hi Alexander:

How does the push request look?

When you integrate it, please let me know the treemacs-version number and then I will test for that in Hyperbole to ensure the right version is utilized.

Alexander-Miller:
> A user-facing function in Emacs is called a command and has an (interactive) form within it. So the two concepts are separate. That is my understanding, anyway.

Before now I haven't had a reason to consider a public api that doesn't consist only out of commands, hence my comment about api being user-facing. Your proposal makes sense. I guess I'll need to commit to not braking that stuff now.

> Mainly, the idea is to get Emacs macros right at compile-time

Sorry, I still don't get it. Let's try a different approach: What difference would it make for hyperbole if loading treemacs was not made inside the eval-and-compile block? I'd like to get this one through my head since it sounds relevant for treemacs, where most of my macros are defined in treemacs-macros.el.

Alexander-Miller:
> How does the push request look?

Just about to the test the code. Don't expect much trouble. Version number will be bumped right after merging.

Alexander-Miller:
Getting quite a few warnings from the compiler, though it looks like many are due to myself being careless in previous commits. Will sort it out tomorrow.

One thing I found already: `treemacs-buffer-exists??` has too many question marks.

rswgnu:
On Tue, Nov 28, 2017 at 4:28 PM, Alexander-Miller <notifications@github.com>
wrote:

> Getting quite a few
> ​​
> ​​
> warnings from the compiler, though it looks like many are due to myself
> being careless in previous commits. Will sort it out tomorrow.
>
> ​Remember you can't know the order in which your package files will be
byte-compiled.
I would recommend you remove whatever larger defsubst you can where they
are not needed for performance reasons as those can make your code more
brittle for little gain.

> One thing I found a
> ​​
> lready: treemacs-buffer-exists?? has too many question marks.
> ​​
>
> ​I will fix that.​

rswgnu:
On Tue, Nov 28, 2017 at 5:29 PM, Robert Weiner <rsw@gnu.org> wrote:

> On Tue, Nov 28, 2017 at 4:28 PM, Alexander-Miller <
> notifications@github.com> wrote:
>
>> One thing I found a
>>
> ​​
>> lready: treemacs-buffer-exists?? has too many question marks.
>> ​​
>>
>> ​I will fix that.​
>

​Done.
​

Alexander-Miller:
Never mind what I said about the warnings. The makefile I use to test compilation is very stupid and managed to load an old treemacs inside my elpa dir instead of the one I'm working on, so I was seeing a lot more, and a lot stranger warnings than there really are.

I've attached a patch for the stuff that should be done. Apply that and squash the commits and we're good to go.
[0001-Warnings-cleanup.txt](https://github.com/Alexander-Miller/treemacs/files/1514823/0001-Warnings-cleanup.txt)

rswgnu:
The rebase and squashing is not working well, giving me all sorts of issues that would take too long to resolve.  Can't you just apply each commit in sequence pretty easily?

Alexander-Miller:
How are you trying to rebase that you're having trouble? Doing an interactive rebase in magit to bring the commits in the right order worked without conflicts for me. I've packed up your changes and pushed them to master now: https://github.com/Alexander-Miller/treemacs/commit/95467c90fa7ca578854d68ad2990b2e12a2aa2aa

I'll bump the version to 1.14 when you confirm that everything is working as intended on your end.

rswgnu:
On Thu, Nov 30, 2017 at 2:29 PM, Alexander-Miller <notifications@github.com>
wrote:

> How are you trying to rebase that you're having trouble?
>
​Magit with 'r' 'i' was dropping one of the commits from the listed ones
above which I had told it to rebase.  It also failed to squash things
properly.
I am on a Mac.

| Doing an interactive rebase in magit to bring the commits in the right
order worked without conflicts for me. I've packed up your changes and
pushed them to master now: 95467c9
<https://github.com/Alexander-Miller/treemacs/commit/95467c90fa7ca578854d68ad2990b2e12a2aa2aa>

Great.

> I'll bump the version to 1.14 when you confirm that everything is working
> as intended on your end.
>
​Okay.

rswgnu:
Testing was looking very good and all features associated with my patch
look good.

Then I came upon this one problem.  It looks like you need to (require
'treemacs-branch-creation) (maybe only eval-when-compile if you like) so
that when the byte-compiler compiles treemacs-impl.el,
treemacs--button-close is defined as a macro.  Everywhere you use macros,
you need to ensure that each file either requires another file that loads
them or load them itself; you can't depend on the requirements being
satisfied by another file like treemacs.el unless you require that since
each file can be byte-compiled in any order.

Otherwise, this error appears:

Debugger entered--Lisp error: (invalid-function treemacs--button-close)
        |;;; treemacs.el --- A tree style file viewer package -*-
lexical-binding: t -*-

  treemacs--button-close(:button #<marker (moves after insertion) at 22 in
*Treem$|

  treemacs--close-node(#<marker (moves after insertion) at 22 in
*Treemacs-Frameb$|;; Copyright (C) 2017 Alexander Miller


  treemacs--push-button(#<marker (moves after insertion) at 22 in
*Treemacs-Frame$|

  treemacs-push-button(nil)
        |;; This program is free software; you can redistribute it and/or
modify

  treemacs-visit-node-default-action(nil)
        |;; it under the terms of the GNU General Public License as
published by

  funcall-interactively(treemacs-visit-node-default-action nil)
        |;; the Free Software Foundation, either version 3 of the License,
or

  call-interactively(treemacs-visit-node-default-action nil nil)
        |;; (at your option) any later version.


  command-execute(treemacs-visit-node-default-action)
        |



​After that fix, everything looks good, so you can move to the next
version.  I did test on a dumb terminal and instead of icons at the left of
each file, I can a small square with a ? mark inside, i.e. missing icons.
Is this something you are working on or should I see unique icons?
​

rswgnu:
Here's an image of the icon problem when running 'emacs -nw':


*hyperbole*                        |              ;; This command requires
recursive minibuffers.


+ .git                             |              (enable-recursive-minibuffers
t))

+ kotl                             |          (unwind-protect

+ man                              |              (progn

- videos                           |                (select-window
(minibuffer-window))

  🗏 hyperbole-hycontrol-20min-vide$|                (set-buffer
(window-buffer (minibuffer-window)))

  🗏 hyperbole-web-search.mov       |                (cond

🗏 .DS_Store                        |                 ;;



🗏 .gitignore                       |                 ;; Selecting a menu
item


🗏 .hypb                            |                 ((eq hargs:reading-p
'hmenu)

🗏 COPYING                          |                  (if assist-flag (setq
hargs:reading-p 'hmenu-help))

🗏 Changes                          |                  (hui:menu-enter
str-value))

🗏 DEMO                             |                 ;;



🗏 DEMO-ROLO.otl                    |                 ;; Enter existing
value into the minibuffer as the desired parameter.


🗏 HY-ABOUT                         |                 ((string-equal
str-value (minibuffer-contents))

🗏 HY-ANNOUNCE                      |                  (exit-minibuffer))

🗏 HY-ANNOUNCE-6.0.2                |                 ;;



🗏 HY-ANNOUNCE-SHORT-6.0.2          |                 ;; Clear minibuffer
and insert value.


🗏 HY-COPY                          |                 (t
(delete-minibuffer-contents)

🗏 HY-NEWS                          |                    (insert str-value)

🗏 HY-WHY.kotl                      |                    (setq back-to t)))

🗏 INSTALL                          |                value)

🗏 MANIFEST                         |            (if back-to (select-window
owind))))))

🗏 Makefile                         |

🗏 README                           |;;;
************************************************************************


🗏 README.md                        |;;; Private variables



🗏 README.md.html                   |;;;
************************************************************************


🗏 TAGS                             |

🗏 TODO                             |(defvar hargs:iforms nil

🗏 TODO-ANNOUNCE-5.15               |  "Alist of (interactive-cmd-chr .
(argument-type . get-argument-form)) elts.")

🗏 TODO-DEFCUSTOMS                  |(setq   hargs:iforms

🗏 TODO-ERRORS                      |        '(

🗏 TODO-HY-ANNOUNCE-6.0.1           |          ;; Get function symbol.



🗏 TODO-MATS-QUES                   |          (?a . (symbol .

🗏 TODO-ORG-MODE                    |                 (intern
(completing-read prompt obarray 'fboundp t default))))

🗏 TODO-ORG-TEST.org                |          ;; Get name of existing
buffer.


🗏 TODO-ORG-vs-HY                   |          (?b . (buffer .

🗏 TODO-Partial-Grep                |                 (progn

🗏 TODO-S-and-P.csv                 |                   (or default (setq
default (other-buffer (current-buffer))))

🗏 TODO-SEQ-README.MD               |                   (read-buffer prompt
default t))))

🗏 TODO-STREAM-README.MD            |          ;; Get name of possibly
nonexistent buffer.


🗏 TODO-hy-5.0.5-changes-to-look-at |          (?B . (buffer .

🗏 _hypb                            |                 (progn

🗏 hact.el                          |                   (or default (setq
default (other-buffer (current-buffer))))

🗏 hactypes.el                      |                   (read-buffer prompt
default nil))))

🗏 hargs.el                         |          ;; Get character.



  ▾ Variables                      |          (?c . (character .

    🞄 hargs:reading-p              |                 (progn (message

    🞄 hargs:iforms                 |                         (if default

    🞄 hargs:iform-vector           |                             (hargs:prompt
prompt

    🞄 hargs:iforms-extensions      |
    (if (integerp default)

    🞄 hargs:iform-extensions-vecto$|
        (char-to-string default)

    🞄 hargs:string-to-complete     |
      default)

  ▸ Functions                      |
    "Curr:")

 Treemacs                          |*s*-UU-:--%--F1  *hargs.el    *   77%
L632,C0   Git-master  (Emacs-Lisp WK Par- Outl Fill) [hargs:select-p]
---------------------------

Alexander-Miller:
treemacs-impl is a general implementation file where I put all the infrastructural bits and pieces that don't clearly belong in one of the dedicated feature files. As such it is only loaded by other parts of the project, but does not load anything other than other "non-loading" modules.
Getting the functions I need there loaded anyway is done with `declare-function`, or mostly the `treemacs--import-functions-from` macro. `treemacs-close-button` is imported in exactly this fashion. It silences the byte compiler, but since you're running into trouble I assume `declare-function` does not work with macros?
If so I'll do some refactoring. `treemacs-close-button` is a good candidate to go into treemacs-branch-creation anyway.

I'm also curious: is there a right way to make sure the compiler knows a macro at the right time, other than explicitly requiring the file the macro is defined in?

rswgnu:
> On Dec 1, 2017, at 5:50 PM, Alexander-Miller <notifications@github.com> wrote:
> , but since you're running into trouble I assume declare-function does not work with macros?
> If so I'll do some refactoring. treemacs-close-button is a good candidate to go into treemacs-branch-creation anyway.
> 

Sounds good.
> I'm also curious: is there a right way to make sure the compiler knows a macro at the right time, other than explicitly requiring the file the macro is defined in?
> 
No, macros are expanded at compile time so they must exist before being used.  You should limit use to just where you need to do code generation.  You can probably change some of your macros into functions if you have a look.

Alexander-Miller:
Pushed an update. Macros should work correctly now.

Alexander-Miller:
Assuming all goes well: I've also pushed version 1.14.

rswgnu:
On Sun, Dec 3, 2017 at 9:46 AM, Alexander-Miller <notifications@github.com>
wrote:

> Assuming all goes well: I've also pushed version 1.14.
>

​I tested it out and it works beautifully.  Nice job.

I came up with one other idea just now (wholly separate) that I'll post as
a separate issue.

Bob

PR #108, opened 2017-11-22
Fix key bindings with use-package

wyuenho:
Recent versions of `use-package` 's `:bind` keyword does not bind the keys after the package has been loaded, the suggested way to bind keys is to use the functions from the `bind-keys` package directly. This commit also fixes some syntax error in the binding list and key chord set up.

### DIFF
--- a/README.org
+++ b/README.org
@@ -236,24 +236,22 @@ You can find an exhaustive overview of all functions, their keybinds and functio
          (treemacs-git-mode 'extended))
         (`(t . _)
          (treemacs-git-mode 'simple))))
-    :bind
-    (:map global-map
-          ([f8]         . treemacs-toggle)
-          ("M-0"        . treemacs-select-window)
-          ("C-c 1"      . treemacs-delete-other-windows)
-          ("M-m ft"     . treemacs-toggle)
-          ("M-m fT"     . treemacs)
-          ("M-m fB"     . treemacs-bookmark)
-          ("M-m f C-t"  . treemacs-find-file)
-          ("M-m f M-t"  . treemacs-find-tag)))
+    :bind (([f8]        . treemacs-toggle)
+           ("M-0"       . treemacs-select-window)
+           ("C-x f 1"   . treemacs-delete-other-windows)
+           ("C-x f t"   . treemacs-toggle)
+           ("C-x f T"   . treemacs)
+           ("C-x f B"   . treemacs-bookmark)
+           ("C-x f C-t" . treemacs-find-file)
+           ("C-x f M-t" . treemacs-find-tag)))
   (use-package treemacs-projectile
     :defer t
     :ensure t
+    :after projectile treemacs
     :config
     (setq treemacs-header-function #'treemacs-projectile-create-header)
-    :bind (:map global-map
-                ("M-m fP" . treemacs-projectile)
-                ("M-m fp" . treemacs-projectile-toggle)))
+    (bind-keys ("C-x f P" . treemacs-projectile)
+               ("C-x f p" . treemacs-projectile-toggle)))
 #+END_SRC
 
 * Configuration


README.org
```diff
@@ -209,24 +209,24 @@ You can find an exhaustive overview of all functions, their keybinds and functio
...
-          ("C-c 1"      . treemacs-delete-other-windows)
-          ("M-m ft"     . treemacs-toggle)
-          ("M-m fT"     . treemacs)
-          ("M-m fB"     . treemacs-bookmark)
-          ("M-m f C-t"  . treemacs-find-file)
-          ("M-m f M-t") . treemacs-find-tag))
+
+    (bind-keys ([f8]         . treemacs-toggle)
+               ("M-0"        . treemacs-select-window)
+               ("C-c 1"      . treemacs-delete-other-windows))
```
Alexander-Miller: The use-package block is missing a closing paren here.

Looks like this error has been here before your change.

README.org
```diff
@@ -209,24 +209,24 @@ You can find an exhaustive overview of all functions, their keybinds and functio
...
     :defer t
     :ensure t
     :config
     (setq treemacs-header-function #'treemacs-projectile-create-header)
-    :bind (:map global-map
-                ("M-m fP" . treemacs-projectile)
-                ("M-m fp" . treemacs-projectile-toggle)))
+    (bind-keys :prefix-map treemacs-prefix-map
+               :prefix "M-m"
+               ("f t"     . treemacs-toggle)
```
Alexander-Miller: Treemacs keybinds should stay with treemacs. These snippets are pick-and-choose and not everyone will care about projectile.

wyuenho:
@Alexander-Miller  See https://github.com/jwiegley/use-package/issues/437#issuecomment-304332850 `:bind` binds the keys regardless of whether the package has been required, so even if `treemacs` hasn't been installed, the key will still be mapped to non-existent commands. Your `:ensure t` will make sure the package is always there, but not every user wants that. By using `(bind-keys)` in the `:config` section, there's only one rule to remember and it does what you'll expect regardless whether the package has been loaded and required.

Alexander-Miller:
I've tried the bind-keys version now and I'm  seeing several issues:
 * A bunch of keybinds don't work. f8 does toggle, but everything under M-m is not bound.
 * For treemacs-projectile nothing is bound.
 * The keybinds are also not run when the package is already present on the system and need not be downloaded.
 * If I don't :ensure the packages then the keybinds will not happen at all, and I need to load the package manually somehow. I'd have expected that a bound key will trigger an installation, but this way I really don't see an advantage. What's the point of a workflow where you set up a package configuration and then don't use it at all?

wyuenho:
You and I are seeing completely different things. I have this on the top my my `.emacs.el` file:

```elisp
(package-initialize)
(require 'use-package)
(require 'bind-key)
```

I also don't use `defer: t` with this set up along with not using `:ensure t`. Can you try again?

> For treemacs-projectile nothing is bound.

This might have because I didn't put `after: treemacs-projectile`, I've just push a commit, can you try again?

> If I don't :ensure the packages then the keybinds will not happen at all, and I need to load the package manually somehow. 

That's expected. That's why I have a `(package-initialize)` statement up top, so all the installed package autoloads will be loaded. A bound key will only trigger an installation if you use :bind, as noted in the issue in my last comment. The point of not using `:ensure t` is that you can control what packages get installed from `package.el`. The point of having a config that doesn't nothing if the package isn't installed is you can just use the same config across different machines with different packages. Sometimes you just want a very lean machine with a minimum amount of packages installed, but if you share the init file, even when you don't have the packages, at least you can some baseline tweaks shared.

If all this is too confusing, it might be a better idea to just give end users the baseline config using `(define-key)` like every other emacs package README does, or better yet, treemacs should define a minor mode that puts all these key bindings in. If the user is using `(use-package)`, they'll all have different ideas about how to use it.

Alexander-Miller:
> I also don't use defer: t with this set up along with not using :ensure t. Can you try again?

That just gets rid of the lazy loading, and loads treemacs right at the start instead of waiting for first use. The keybind situation is back to where it was, with the M-m binds not working.

> This might have because I didn't put after: treemacs-projectile, I've just push a commit, can you try again?

Working correctly now.

> You and I are seeing completely different things. I have this on the top my my .emacs.el file:

Spacemacs is managing my packages. So long as the treemacs layer is not part of stable spacemacs the install snippet must work there as well.

> The point of having a config that doesn't nothing if the package isn't installed is you can just use the same config across different machines with different packages. 

This forces you to manually install all your packages instead of sticking a bunch of :disabled keywords into your config, or hiding the maybe-I-don't-need-it-stuff behind a `when` call. Far as I can see this is all your own personalized workflow that offers no improvement for the general use-case of the install code. That install snippet is supposed to be a general example where you cherrypick the parts you want. Adaption to a scenario where you configure but never use treemacs is outside its scope.

> If all this is too confusing, it might be a better idea to just give end users the baseline config using (define-key) like every other emacs package README does,

Treemacs is not every other package. As far as documentation goes I like to err on the side of being thourough. By now treemacs is firmly in the territory of having so many options that it's overwhelming to new users and 90% of them will probably go unused. I like to think that by offering an install snippet that configures treemacs to its defaults will encourage some exploration.

You're right that everyone will have their own ideas of how to use use-package, and that is exactly why I want my install snippet to be comprehensive - it's so that people will have all the building blocks they need for their preferred setup.

wyuenho:
> That just gets rid of the lazy loading, and loads treemacs right at the start instead of waiting for first use. The keybind situation is back to where it was, with the M-m binds not working.

Are you saying the key binds for `projectile-treemacs` but not those under `treemac`'s `:config` section? You said `[f8]` works but all the other `M-m` keys don't, but then the `projectile-treemacs` key-binding works. This makes no sense, either all the M-m keys works or none of them do.

I think if you want to be *general*, you should provide configuration example that works on vanilla emacs, i.e. `(define-keys)`. If you want to be *comprehensive*, you should provide configuration example for `(use-package)` *that works* on vanilla emacs, and possibly a configuration for spacemacs, which I don't use, but AFAIK, has its own convention. As of now, you've provided no configuration example for vanilla emacs, and your `(use-package)` doesn't work on vanilla emacs without defining a prefix map first like in this PR. After all, M-m is a very unusual prefix map that few will have defined before reaching the treemacs block.

Lastly, I suspect your spacemacs config might have defined or redefined the M-m leader keymap somewhere, that might explain the weird behavior you are seeing.

Alexander-Miller:
> This makes no sense, either all the M-m keys works or none of them do.

It's due to spacemacs. In the absence of treemacs most of those M-m keybinds are taken by neotree. Using :bind would overwrite them, bind-keys does not.

> and your (use-package) doesn't work on vanilla emacs without defining a prefix map

Ah, I did not know that. For now no solution is ideal, current version would not work in vanilla, yours would not work with spacemacs. 
I'm the one to maintain the installation snippet, so my preference goes to whatever version is less likely to cause trouble later. IMO that would mean providing an explanation for the prefix issue and maybe also change the keybinds to `C-c f t` etc. This prefix is meant for  user bindings and is free on both vanilla and spacemacs.

wyuenho:
@Alexander-Miller I've bound the suggested keybindings to `ctl-x-map`. `C-c` belongs to `mode-specific-map`, which is generally quite crowded and may get clobbered by other modes. Since `C-x t` is pretty much the only thing available on `ctl-x-map`, let take advantage of that.

wyuenho:
How does this look now?

Alexander-Miller:
Getting a conclusion here is next on my list.
Let's see:
 * It's all sorts of annoying to deviate from the spacemacs setup with `C-x t`, but if nothing else is available then let's go with that.
 *  The `bind-keys` use in projectile needs to be turned back to `:bind`. Right now it's part of `:config` so those keybinds are only made *after* treemacs-projectile has loaded.
 * The conflicts should be minor. I've added some config variables and put the `M-0` bind for winum into `:init` so it's loaded at the right time.

I believe that's all.

Alexander-Miller:
I've updated the installation snippet for the C-x t prefix and for the new v2 changes. Everything's working fine in my tests, so this PR should no longer be needed.

PR #113, opened 2017-11-29
Add a configuration variable to control whether linum-mode will be enabled

louietan:


Alexander-Miller:
Gonna have to refuse this one, for two reasons:

 1. There are different line number implementations. Linum is builtin and very slow, nlinum, improves that and emacs 26 has native line numbering implemented in C. Ideally you'd configure all that as well, which brings us to the next point:
 2. You've basically added a roundabout mode hook. The canonical way to activate a minor mode mode in certain buffers is to add that minor mode activation function to the appropriate mode hook. So to do what you want is a simple one-liner: `(add-hook 'treemacs-mode-hook #'(n)linum-mode)`, best added to the `:config` part of use-package. You can use whatever line number implementation or any other minor modes you want and there is no need to make changes in treemacs for special cases.

Issue #17, opened 2017-06-16
`shell-pop-eshell` fails when treemacs buffer is open


TimoFreiberg:
error message: `split-window: Cannot split side window or parent of side windowInvalid face reference: spacemacs-treemacs-face`

Alexander-Miller:
works for me :(

SPC t D + stacktrace pls

TimoFreiberg:
```
Debugger entered--Lisp error: (error "Cannot split side window or parent of side window")
  signal(error ("Cannot split side window or parent of side window"))
  error("Cannot split side window or parent of side window")
  split-window(#<window 11> 33 below)
  shell-pop-split-window()
  shell-pop-up(1)
  shell-pop(nil)
  (if (equal (quote (4)) index) (eshell nil) (shell-pop--set-shell-type (quote shell-pop-shell-type) (backquote ("eshell" "*eshell*" (lambda nil (eshell nil))))) (shell-pop index))
  spacemacs/shell-pop-eshell(nil)
  funcall-interactively(spacemacs/shell-pop-eshell nil)
  call-interactively(spacemacs/shell-pop-eshell nil nil)
  command-execute(spacemacs/shell-pop-eshell)
```

ambihelical:
I can reproduce this shell-pop w/o spacemacs, but it only happens when treemacs has the cursor.  
But isn't this just a limitation of any side window though?

Anntoin:
Also seeing this behaviour, same error. Was just trying the treemacs layer in the hopes this would work
For reference here's the neotree issue https://github.com/syl20bnr/spacemacs/issues/7446
Would be great if treemacs was able to fix this

Alexander-Miller:
@ambihelical is right - it's a general side window limitation. I am currently busy making sure treemacs' side window status is kept intact when using eyebrowse and persp (though this issue is making me doubt whether it should have it in the first place). When this is done I should have the means I need to fix this issue.

In the meantime you can use eyebrowse as a workaround - switch to a different workspace and back and treemacs loses its side window status, which means that 1) the treemacs window gets closed by `delete-other-windows` or `magit-status` and 2) functions like pop-shell-* work again.

Anntoin:
@Alexander-Miller thanks for the info, understanding the side window problem led me to another workaround. For me I just now  set `shell-default-full-span nil` so the terminal doesn't split the side window. For my usage this is perfect and works for treemacs and neotree.

Alexander-Miller:
[...]

PR #117, opened 2017-12-05
New options to display Treemacs on the right and to preserve Treemacs window from delete-other-windows

rswgnu:
Hi Alexander:

This pull request is just for the latest commit 9a2bb2f.  Please ignore the other commits showing.

I know you were exploring using side-windows in another Treemacs branch (as this pull request does too) but I wanted something fairly simple that would provide an option to prevent delete-other-windows from deleting any Treemacs window.  And since it was easy to do at the same time, I added an option to allow for placement of Treemacs windows on the right.  The defaults of both new options leave Treemacs behavior the same as it was; you have to set them to see any change and if you set them with customize-set-variable, the change is dynamic and will affect all existing and future Treemacs windows.  Give it a try and let me know what you think.


treemacs-customization.el
```diff
@@ -70,6 +70,45 @@ times."
...
 
+(defun treemacs-windows (&optional frame)
+  "Return a list of Treemacs windows across all frames or just from optional FRAME."
+  (delq nil (mapcar (lambda (win)
+		      (when (eq (buffer-local-value 'major-mode (window-buffer win))
+				'treemacs-mode)
+			win))
+		    (window-list-1 nil nil (or frame t)))))
+
+(defcustom treemacs-window-location 'left
```
Alexander-Miller: That's already available under `treemacs-position`. Without the extra set option, but I don't imagine it's something that'll change often.

treemacs-impl.el
```diff
@@ -684,15 +684,18 @@ treemacs buffer teardown\) otherwise the currently selected frame is used."
...
-      (frame-root-window)
-      (split-window nil treemacs-position)
-      (select-window))
-  (let ((buf (treemacs--get-framelocal-buffer)))
-    (switch-to-buffer buf)
-    (bury-buffer buf))
-  (set-window-dedicated-p (selected-window) t)
-  (let ((window-size-fixed))
+  (let* ((buf (treemacs--get-framelocal-buffer))
+	 (win (select-window (display-buffer-in-side-window
```
Alexander-Miller: No side windows, at least not without an option. 1) They are unsplittable, which may leads to conflicts with packages like shell-pop. 2) The window parameters that define a side window have been changed with emacs26, so that would have to be taken into account as well.

Best just stick with slitting the frame root.

rswgnu:
On Wed, Dec 6, 2017 at 12:29 PM, Alexander-Miller <notifications@github.com>
wrote:

> *@Alexander-Miller* commented on this pull request.
> ------------------------------
>
> In treemacs-impl.el
> <https://github.com/Alexander-Miller/treemacs/pull/117#discussion_r155304566>
> :
>
> No side windows, at least not without an option. 1) They are unsplittable,
> which may leads to conflicts with packages like shell-pop.
>
​That was the problem I was trying to solve.  display-buffer splits a
Treemacs window and displays things underneath the treemacs window.  If
Treemacs is 20 characters wide, I see a file in a narrow window like that,
which is undesirable.  So I think it's windows should be unsplittable by
default.  Why don't we look at what the issue is with shell-pop.  Can you
explain it?​

My suggestion is to try using it this way and see what issues you can find
and then discuss them.
​​​

> ​​
> 2) The window parameters that define a side window have been changed with
> emacs26, so that would have to be taken into acco
> ​​
> unt as well.
>
> ​Ok, I could look at that and fix it.

I'm not averse to doing this via the root window but we'd need another way
to make it unsplittable, e.g. split-*-threshold.

Bob
​

rswgnu:
On Wed, Dec 6, 2017 at 12:13 PM, Alexander-Miller <notifications@github.com>
wrote:

> T
> ​​
> hat's already available under treemacs-position. Without the extra set
> option, but I don't imagine it's something that'll change often.
> ​​
> ​​
>
>
​I didn't see that before, thanks.
​
​Why is that option only used in the treemacs--execute-button-action
macro?  How is it applied when a treemacs window is first created.
You could just add the :set logic I provided to this option if it indeed is
used whenever a new Treemacs window is created.

Bob

Alexander-Miller:
> Why is that option only used in the treemacs--execute-button-action macro? 

It isn't. Check your diffs, it's also used in `setup-buffer`.

> You could just add the :set logic I provided to this option if it indeed is used whenever a new Treemacs window is created. 

The `:set` hook only applies when the value is changed through the customize interface, right? If so I prefer it'd be left out. Just about every single customization option only applies after the next time the treemacs view/window/buffer is rebuilt. I'd rather not have to apply some type of refresh after every change to keep things consistent.

> Why don't we look at what the issue is with shell-pop. Can you explain it?​

See https://github.com/Alexander-Miller/treemacs/issues/17. Last comment has more links. To see for yourself: install shell-pop and try to call it when point is in a side-windowed treemacs buffer.

> If Treemacs is 20 characters wide, I see a file in a narrow window like that, which is undesirable.

In what situation does that happen? Do you use something like popwin or shackle to manage your windows?

> I'm not averse to doing this via the root window but we'd need another way to make it unsplittable, e.g. split-*-threshold.

Those are for `split-window-sensibly`. When is it used? Changing the values for `C-x 2` does nothing,

PR #122, opened 2017-12-15
Use unicode bullet instead of black slightly small circle

wyuenho:
The `treemacs-icon-tag-leaf-text` now is the unicode black slightly small circle and it's quite rare as a glyph in many fonts. I suggest you switch to using the very similar but far more common unicode bullet instead.


Alexander-Miller:
👍

Issue #79, opened 2017-10-13
[Discuss] The behavior of mouse-click


seagle0128:
While clicking on a file without any tags in treemacs, a message pops up like this:
```elisp
[Treemacs] No imenu index found for /Users/vincent/.dotfiles/LICENSE
```
The behavior is same as `TAB`. I think the better behavior is opening the file instead, like `RET`.  It's more convenient to visit the files quickly. I am struggling to `Click` or `TAB` without response but only a message, and have to press `RET`. Your opinion?

Alexander-Miller:
Good timing. I have just finished a proof of concept for doing double clicks, so weve twice the playroom now.

Heres my propostion: We leave single clicks as it is, as a simple expand/close mechanism. Double clicks will work the same as RET, so theyll open files as youre requesting. Maybe I can add some minor changes, for example a double click on a directory could still just expand a directory, since this is the way its done in IDEs like eclipse & intellij.
Additionally doube clicks can be configured via a mapping like RET is with `treemacs-default-actions`.

What do you think?
Also asking you, @TimoFreiberg, @bitterblue.

TimoFreiberg:
I'm voting for double click == `RET`

seagle0128:
Good idea! But I prefer click == `RET` and double click == `TAB`.
Maybe adding a customization is a better way :-)

bitterblue:
Not much of a mouse user, so no strong opinions here. @TimoFreiberg's preference makes it easier to explore directory structures, @seagle0128's makes it easier to quickly browse through files. Both is fine with me.

Alexander-Miller:
Configuration it is then. Ill see to it that I come up with some sane defaults, but youll get adjust things as you please.
Customization will be done the same way as RET and `treemacs-default-actions`, so best take a look at its docs.

TimoFreiberg:
After thinking about it and testing what `RET` actually does with directories, I agree with @seagle0128 !

rswgnu:
I have added Treemacs mouse support to Hyperbole's next release.  For users that enable unshifted-mouse key support, in a Treemacs buffer, mouse-2 will perform the RET function with a single click on the item label/name and the TAB function with a single click on the item's icon.  This is simpler and clearer than moving to double clicks, I believe, because you have a separate visual item to target for each different function.  Any interest in doing it that way?

wyuenho:
> click == RET and double click == TAB. It seems more conventional

[...]

PR #123, opened 2017-12-15
open file on double click

wyuenho:
Currently double clicking on a file in treemacs will only attempt to open the file's tags. If no tags are found, there's no way to visit a file with a mouse click. This PR fixes it.


Alexander-Miller:
As per the discussion in https://github.com/Alexander-Miller/treemacs/issues/79 most people wanted to go the other way: click = open file, double-click = expand content


Your solution will also not work correctly since the mouse1 function is called with both a single and double click. In general a double-click is not supposed to do more than a single click, but that's not what we're doing in treemacs, so make this work clicks would have to work with timers.

wyuenho:
If anything this change is inconsistent, since the "open" action is done using both a single click and a double click, but this actually will work. The double click action is debounced by emacs already, you are not going to receive 2 single and 1 double.

wyuenho:
I can make this configurable if you want.

wyuenho:
Closing this, I'm gonna try something different. tab/enter/mouse-click are just too weird in treemacs now.

PR #124, opened 2017-12-16
Temporarily disable run-mode-hooks during imenu index

wyuenho:
People set all kinds of mode hooks to the major modes. For modes such as js-mode or python-mode, there are a lot of linters or code analyzers that call out to external processes. The problem currently with treemacs's way to get tags for a file is that it uses a real buffer complete with it's major mode, and as soon as it has gathered all the tags from the imenu index, the buffer is killed immediately, while the launched processes are still running. When the processes return a result, the buffer would have been killed and a `Selecting deleted buffer` error will result, and stops the rest of the get tags business. This PR fixes this bug. This PR also makes getting the tags slightly faster as well.

### DIFF
--- a/src/elisp/treemacs-tags.el
+++ b/src/elisp/treemacs-tags.el
@@ -139,18 +139,18 @@ should be placed under."
         (existing-buffer (get-file-buffer file)))
     (if existing-buffer
         (setq buff existing-buffer)
-      (setq buff (find-file-noselect file)))
-    (with-current-buffer buff
-      (setq result (condition-case _
-                       (imenu--make-index-alist t)
-                     (error nil))
-            mode major-mode))
-    (unless existing-buffer (kill-buffer buff))
-    (when result
-      (when (string= "*Rescan*" (car (car result)))
-        (setq result (cdr result)))
-      (unless (equal result '(nil))
-        (treemacs--post-process-index result mode)))))
+      (letf (((symbol-function 'run-mode-hooks) (symbol-function 'ignore)))
+        (setq buff (find-file-noselect file))))
+    (when (buffer-live-p buff)
+      (with-current-buffer buff
+        (setq result (imenu--make-index-alist t)
+              mode major-mode))
+      (unless existing-buffer (kill-buffer buff))
+      (when result
+        (when (string= "*Rescan*" (car (car result)))
+          (setq result (cdr result)))
+        (unless (equal result '(nil))
+          (treemacs--post-process-index result mode))))))
 
 (defun treemacs--add-to-tags-cache (btn)
   "Add BTN's path to the cache of open nodes."


src/elisp/treemacs-tags.el
```diff
@@ -136,21 +136,25 @@ should be placed under."
   (let ((buff)
         (result)
         (mode)
+        (run-mode-hooks-override (lambda (&rest _) nil))
```
Alexander-Miller: Don't need this, you can just use `ignore`.

src/elisp/treemacs-tags.el
```diff
@@ -136,21 +136,25 @@ should be placed under."
...
-                     (error nil))
-            mode major-mode))
-    (unless existing-buffer (kill-buffer buff))
-    (when result
-      (when (string= "*Rescan*" (car (car result)))
-        (setq result (cdr result)))
-      (unless (equal result '(nil))
-        (treemacs--post-process-index result mode)))))
+      (progn
+        (advice-add 'run-mode-hooks :override run-mode-hooks-override)
```
Alexander-Miller: That's a good case for `cl-flet`:
`(cl-flet ((run-mode-hooks (&rest _) (ignore))) ... )`

Alexander-Miller:
👍 

> The problem currently with treemacs's way to get tags for a file is that it uses a real buffer complete with it's major mode

If there's another way to get at imenu's stuff then I'm all ears.

> and as soon as it has gathered all the tags from the imenu index, the buffer is killed immediately

Only if that file wasn't open already.

wyuenho:
> If there's another way to get at imenu's stuff then I'm all ears.

Give `semantic` another half a decode :)

Alexander-Miller:
We'll probably be running on language servers by then.

wyuenho:
@Alexander-Miller how does this look now?

Alexander-Miller:
Still needs to use cl-flet. Advice is the wrong tool for this kind of quick temp-replace job.

wyuenho:
@Alexander-Miller that doesn't work. `cl-flet` is l[exically bound, the references have to be within the body](https://www.gnu.org/software/emacs/manual/html_node/cl/Function-Bindings.html), but I'm temporarily monkey-patching the global reference here.

Alexander-Miller:
Ah, now that's something I should remember.

Advice is still quite heavyweight, so I think the next best solution is to directly switch the symbol's function value. Grab `run-mode-hooks`' function value with `symbol-function`, replace it with `ignore` with `fset`, then get imenu and put the original back.

wyuenho:
Using good old `letf` now, seems to work pretty well.

Alexander-Miller:
> letf

I didn't know you could use something like this with functions. That's even better than my own idea 👍 
I'll be merging this this now.

PR #155, opened 2018-03-06
Add .tsx extension for typescript and add vue.js icon for .vue files

alanwflood:
Typescript uses .tsx as an extension for using typescript syntax in jsx.

.vue extension is used by the vue.js javascript framework.


Alexander-Miller:
Thanks :+1:

Issue #158, opened 2018-03-11
Support option to sort alphabetically without taking capitalization in to consideration


fiveNinePlusR:
It would be nice to have an option that downcases every file before deciding on sort order. This would fix the problem where capitalized folders appear higher in the list than lower case folders/files... effectively splitting each folder into four groups of display: 

1. capitalized folders 
2. lowercase folders
2. capitalized files
2. lower case files

which is confusing in general. I know we talked about the performance repercussion of this so it would need to be an alternative sort option and not set as the default.  I will see if I can get a patch for this feature soon.

Alexander-Miller:
Adding that option should be easy enough. Your entry point is [over here](https://github.com/Alexander-Miller/treemacs/blob/master/src/elisp/treemacs-customization.el#L189). You'd just need to add `alphabetic-asc/desc-ignore-case` options.
If you need help getting around the code base I am happy to help.

Alexander-Miller:
To answer your questions from gitter:

 * Don't know about pfuture. I just tried cask install after deleting .cask and run into `Variable binding depth exceeds max-specpdl-size`, which seems to be an issue with shut-up.el. It did work after trying a second time. So try again I guess.

 * No tests for sorting. I'll put it on my todo list to add some to the systest, but that's only after multiroot is done, since the sys test will need to be largely rewritten anyway.

fiveNinePlusR:
I did get cask working after some fiddling about. all seems good now.

PR #160, opened 2018-03-12
Fixes: #158 - Add case insensitive sort options.

fiveNinePlusR:
The two options for sorting:

* `alphabetic-case-insensitive-asc`
* `alphabetic-case-insensitive-desc`


Alexander-Miller:
> All checks have failed

Do you happen to know your way around travis? I'm pretty sure my emacs26 builds are allowed to fail, but the setting is not applied.

src/elisp/treemacs-branch-creation.el
```diff
@@ -43,13 +43,21 @@ is a marker pointing to POS."
   (copy-marker pos t))
 
 (defsubst treemacs--sort-alphabetic-asc (f1 f2)
-  "Sort F1 and F2 alphabetically asc."
+  "Sort f1 and f2 alphabetically asc."
```
Alexander-Miller: Function params must be capitalized or the compiler complains. `make test` can catch these (this stuff will go into a Contributing readme section).

src/elisp/treemacs-branch-creation.el
```diff
@@ -99,6 +107,8 @@ DEFAULT: Face"
           (-pcase treemacs-sorting
             [`alphabetic-asc  #'treemacs--sort-alphabetic-asc]
             [`alphabetic-desc #'treemacs--sort-alphabetic-desc]
+            [`alphabetic-case-insensitive-asc  #'treemacs--sort-alphabetic-case-insensitive-asc]
+            [`alphabetic-case-insensitive-desc #'treemacs--sort-alphabetic-case-insensitive-desc]
             [`size-asc        #'treemacs--sort-size-asc]
```
Alexander-Miller: You should also delete the extra whitespace in the other match arms. I try to keep things nicely aligned, but not if the difference is as large as with the new values.

fiveNinePlusR:
I updated my branch to allow failures in emacs 26

fiveNinePlusR:
Is a rebase and squash something that you'd like to see happen to clean up the history?

Alexander-Miller:
Yes, so long as the changes are small enough there should be one commit per topic, so the indent update should be squashed.

And thanks for setting travis straight.

fiveNinePlusR:
I am not sure how to mark that the changes have been made to the commits inside github but it's been updated.

Alexander-Miller:
I don't think github has that feature. At any rate we're good to merge :+1:

PR #163, opened 2018-03-20
An option to customize the icon size.

amfranz:
The icons aren't quite the right size for me.

I mainly work with two environments, in one of them I use a 200 DPI monitor and therefore I need the icons to be double the size to match the size of the text.

The other environment I work with has a 96 DPI monitor but I use a small font, which means I actually need the icons to shrink a little to match the size of the text.

This variable allows me to right-size the icons to accomodate both.

### DIFF
--- a/src/elisp/treemacs-visuals.el
+++ b/src/elisp/treemacs-visuals.el
@@ -62,6 +62,10 @@
 Used by `treemacs--setup-icon-highlight' to realign icons' highlight colors
 after a theme change.")
 
+(defvar treemacs--icon-size nil
+  "Size in pixels icons will be resized to.
+See also `treemacs-resize-icons'.")
+
 (defvar-local treemacs--last-highlight nil
   "The last button treemacs has highlighted.")
 
@@ -143,11 +147,18 @@ argument."
       (error
        (treemacs-log "Error on highlight, this shouldn't happen: %s" e)))))
 
+(defun treemacs--create-image (file-path)
+  "Load image from FILE-PATH and resize it to the configured icon size."
+  (if (and (integerp treemacs--icon-size) (image-type-available-p 'imagemagick))
+      (create-image file-path 'imagemagick nil :ascent 'center
+                    :width treemacs--icon-size :height treemacs--icon-size)
+    (create-image file-path 'png nil :ascent 'center)))
+
 (defmacro treemacs--setup-icon (var file-name &rest extensions)
   "Define string VAR with its display being the image created from FILE-NAME.
 Insert VAR into icon-cache for each of the given file EXTENSIONS."
-  `(let* ((image-unselected (create-image (f-join treemacs-dir "icons/" ,file-name) 'png nil :ascent 'center))
-          (image-selected   (create-image (f-join treemacs-dir "icons/" ,file-name) 'png nil :ascent 'center)))
+  `(let* ((image-unselected (treemacs--create-image (f-join treemacs-dir "icons/" ,file-name)))
+          (image-selected   (treemacs--create-image (f-join treemacs-dir "icons/" ,file-name))))
      (treemacs--set-img-property image-selected   :background treemacs--selected-icon-background)
      (treemacs--set-img-property image-unselected :background treemacs--not-selected-icon-background)
      (defconst ,var
@@ -256,7 +267,28 @@ inserted into `treemacs-icons-hash'."
           treemacs-icon-tag-leaf-text        (alist-get 'treemacs-icon-tag-leaf-text        treemacs--defaults-icons)))
   (treemacs--create-icons)
   (when (fboundp 'clear-image-cache)
-    (clear-image-cache)))
+    (clear-image-cache t))
+  (treemacs-run-in-every-buffer
+   ;; the reason this is called is because it updates the buffer local variables
+   ;; `treemacs-icon-open', `treemacs-icon-closed', etc.
+   (treemacs--check-window-system)
+   (treemacs--do-refresh (current-buffer))))
+
+;;;###autoload
+(defun treemacs-resize-icons (size)
+  "SIZE in pixels icons should be resized to.
+
+If SIZE is `nil' the icons are not resized and will retain their default size of
+22 pixels.
+
+There is only one size, the icons are square and the aspect ratio will be
+preserved when resizing them therefore width and height are the same.
+
+Resizing the icons only works if Emacs was built with ImageMagick support. If
+this is not the case this function will report an error."
+  (interactive "nIcon size in pixels: ")
+  (setq treemacs--icon-size (if (= size 22) nil size))
+  (treemacs-reset-icons))
 
 (defun treemacs-map-icons-with-auto-mode-alist (extensions mode-icon-alist)
   "Remaps icons for EXTENSIONS according to `auto-mode-alist'.
--- a/README.org
+++ b/README.org
@@ -287,6 +287,10 @@ You can find an exhaustive overview of all functions, their keybinds and functio
             treemacs-tag-follow-delay           1.5
             treemacs-width                      35)
 
+      ;; The default width and height of the icons is 22 pixels. If you are
+      ;; using a Hi-DPI display, uncomment this to double the icon size.
+      ;;(treemacs-resize-icons 44)
+
       (treemacs-follow-mode t)
       (treemacs-filewatch-mode t)
       (pcase (cons (not (null (executable-find "git")))


src/elisp/treemacs-visuals.el
```diff
@@ -140,11 +140,18 @@ argument."
       (error
        (treemacs-log "Error on highlight, this shouldn't happen: %s" e)))))
 
+(defun treemacs--create-image (file-path)
+  "Load image from FILE-PATH and resize it to the configured icon size."
+  (if (and treemacs-resize-icons (image-type-available-p 'imagemagick))
```
Alexander-Miller: To be really safe you should use `(integerp treemacs-resize-icons)`.

Alexander-Miller:
Bump?

src/elisp/treemacs-visuals.el
amfranz: Good catch, I'll add that.

amfranz:
I apologize for the delay.

I added a change to `treemacs-reset-icons` which refreshes all open treemacs buffers, so that a change in icon size becomes visible in existing treemacs buffers.

I have a few concerns about the pull request as a whole:

The logic to resize the icons is working, but its configuration comes with a caveate. Customizing the variable `treemacs-resize-icons` does not take effect when it is set in the `:config` block of `use-package`. At that point the package has been loaded already, therefore the icon cache has already been populated with icons of the default size.

To work around that one has to either set `treemacs-resize-icons` in the `:init` block, or alternatively call `treemacs-reset-icons` in the `:config` block somewhere after setting the variable. Both of these approaches have the same problem: they are not obvious, they violate the principle of least surprise.

To mitigate that, I think the loading of the icons should be deferred until the first treemacs buffer is opened. This way, `treemacs-resize-icons` can be configured in the `:config` block and will still take effect, because the icons have not been loaded yet.

Do you think it's a good idea to do go down this road? Do you think loading the icons late causes any issues? If the icons can be loaded late, where do you think would be a good place to put the call that loads them, maybe in `treemacs--init`?

src/elisp/treemacs-visuals.el
```diff
@@ -240,7 +247,12 @@ inserted into `treemacs-icons-hash'."
           treemacs-icon-tag-leaf-text        (cdr (assq 'treemacs-icon-tag-leaf-text        treemacs--defaults-icons))))
   (treemacs--create-icons)
   (when (fboundp 'clear-image-cache)
-    (clear-image-cache)))
+    (clear-image-cache t))
+  (dolist (buf (buffer-list))
+    (with-current-buffer buf
+      (when (eq 'treemacs-mode major-mode)
+        (treemacs--check-window-system)
```
Alexander-Miller: That call isn't needed here. That check only needs to be made when a treemacs buffer is first created.

src/elisp/treemacs-visuals.el
```diff
@@ -240,7 +247,12 @@ inserted into `treemacs-icons-hash'."
           treemacs-icon-tag-leaf-text        (cdr (assq 'treemacs-icon-tag-leaf-text        treemacs--defaults-icons))))
   (treemacs--create-icons)
   (when (fboundp 'clear-image-cache)
-    (clear-image-cache)))
+    (clear-image-cache t))
+  (dolist (buf (buffer-list))
```
Alexander-Miller: No need to iterate over the whole list. You'll find a frame -> buffer alist in `treemacs--buffer-access`.

Alexander-Miller:
> Do you think it's a good idea to do go down this road?

Yes, you raise good points. A good place for lazy icon creation would be the treemacs-mode initialization code. The current call to create-icons is already in that file and would just need to be moved. 
Of course the icon creation needs to only ever run once. For this I can't think of a nicer solution than
```
(unless (boundp 'treemacs--icons-created)
  (defvar treemacs--icons-created t
    "Defined when treemacs' icons have been created.")
  (treemacs--create-icons))
```

src/elisp/treemacs-visuals.el
amfranz: That function makes some icon variables buffer-local, basically keeping (some of) the old icons bound to the buffer.

The reason I call it again is to reset those buffer-local variables, to force the buffer to adopt the new icons.

Without this call, existing Treemacs buffers will retain (most of) their old icons even after refreshing the buffer.

I know where you are coming from, calling this again does feel awkward to me.

Architecturally, I question making the variables buffer-local. They are set in `(treemacs--create-icons)` with `setq`, which means this function sets either their local or global value depending on which buffer is active when the function gets called (it probably should use `setq-default`). That also applies to `treemacs-reset-icons` since it calls `treemacs--create-icons`.

I didn't think this through, but maybe testing for `(display-graphic-p)` at the time of rendering the icons would be logic that is easier to reason with, though it would likely be less performant.

Alexander-Miller: With icon creation you've gotten involved with pretty much the ugliest part of the code base.
 * The declaration and definition order is awkward since I need to take into account `treemacs--image-creation-impossible`, but the icons should also be defined at the top level so the compiler knows those variables when requiring that module.
 * `check-window-system` is mostly a remnant of the time before treemacs buffers were frame-scoped, I had to call it many more times than I do now.
 * Being able to restore icons adds another layer on complexity.
 * Some variables also do need to be buffer-local so treemacs can correctly switch them when it detects a buffer is running in a terminal. From I you say I need to pay closer attention to which variables are global/local when making those changes, too.
 * Shifting work to render is not something I'll do unless I have no choice, and so far the system works, even if it's not the cleanest.

And now we want to complicate things some by adding lazy initialization, though I suppose it's a good opportunity to do some house keeping. Now this is quite the hornet's nest and well beyond the scope of your PR, so it's all up to you if you want to go along with all those refactoring plans. I'm fine with merging this with the few changes that need to be made now and cleaning things up later.

src/elisp/treemacs-visuals.el
amfranz: Fantastic. `treemacs-run-in-every-buffer` makes it a lot easier, indeed. I changed the logic to use it.

src/elisp/treemacs-visuals.el
amfranz: Thank you for the insights into the evolution of the code, I think I understand the level of complexity involved better now.

> And now we want to complicate things some by adding lazy initialization, though I suppose it's a good opportunity to do some house keeping. Now this is quite the hornet's nest and well beyond the scope of your PR, so it's all up to you if you want to go along with all those refactoring plans. I'm fine with merging this with the few changes that need to be made now and cleaning things up later.

Well, my aim is not to complicate, quite the opposite. Let's take a step back. This pull request is just meant to make it possible to resize the icons, so let's focus on that. Whether to implement lazy initialization or do house-keeping is something that can be tackled later.

The pull request as-is fulfills the original purpose of making it possible to change the icon size, just with one caveate: if `treemacs` has been loaded already, it is necessary to not only change the value of `treemacs-resize-icons`, but also to call `(treemacs-reset-icons)` to re-load the icon cache. Most users will customize settings in `use-package`'s `:config` section, so `treemacs` will have been loaded already.

My suggestion is to accept this caveate for now, because accepting this caveate is much easier than a re-work of the icon initialization logic. I really don't think I have sufficient time available to commit to that.

I could just document the caveate, in the `defcustom treemacs-resize-icons` help text and the `README`.

Consequently calling `treemacs-reset-icons` would also be have to worked into the usage example in the `README`, since that one shows the variable that can be customized. But I imagine would be a bit awkward to explain why/when you need to call it: only if `treemacs-reset-icons` is not `nil`.

Now that I think about it, to make the usage example less awkward, in my opinion a viable option would be to replace the `defcustom treemacs-resize-icons` with a function of the same name that does all the necessary steps: updates a package-private `defvar` to remember the new icon size, and calls `treemacs-reset-icons` afterwards. Then all that would be needed in regards for documentation is to document the existence of the function, and show how it can be called in the usage example. It makes the documentation simpler because there is no remind users to call `treemacs-reset-icons` themselves, nor can users forget to call it.

Alexander-Miller: > If you are ok with this as a solution, I can merge the changes from that branch into this pull request.

I think an interactive function is an even better solution for this use-case than a def-custom since a function allows for interactive changes on the fly. It also opens up the option to later create functions that will interactively scale the icons by single pixels.

Alexander-Miller:
Bump! Where did we leave this off? I tried out your branch a couple days ago and every seemed to work fine. Looks like you just need to rebase on the newest master and squash your changes and we can merge this, or am I forgetting something?

I'll go over our conversation later and check which parts I should extract as issues.

amfranz:
Yes, you are right, it works fine now. I apologize for the delay.

I changed the method to customize the icon size to be an interactive `defun` instead of a `defcustom`, and documented it as such in the README.

This defun then takes care of doing the call to `treemacs-reset-icons` which currently is necessary so the icon cache gets cleared and existing treemacs buffers are redrawn.

I squashed the branch and rebased it onto the latest master as requested.

So this PR is ready to merge now, in my opinion. You've got the final word on this.

Alexander-Miller:
:+1:

Alexander-Miller:
FYI: There's some bugs here after all. I need to create a workaround for the project root icon, since it's not actually square. The non-default file icons are also not changing their size.

I've a good idea on where this is coming from, so these bits will be fixed soon.

PR #186, opened 2018-05-06
[FIX] treemacs-select-window init treemacs buffer

mpanarin:
If treemacs buffer was not initialized beforehand - treemacs-select-window will be broken. As it is bound to `M-m 0` it could be used as general *select or open treemacs* key.
So it should check if it exists and init it, if its not


mpanarin:
@Alexander-Miller oops, seems like you added the fix, sry

Alexander-Miller:
¯\\\_(ツ)\_/¯

Issue #183, opened 2018-05-05
`treemacs` command clarification - should it toggle?


xdl:
When trying to use`treemacs` as the new replacement for `treemacs-toggle` (as its description is `Show/Hide/Initialize treemacs`), I noticed it doesn't hide the treemacs pane, only shows or initialises it - is this a bug or does the readme need to be updated? Either way, it would be nice to have some sort of toggling command exposed again!

Alexander-Miller:
That's a bug, looks like I misused `bury-buffer`. Patch is pushed now.

xdl:
Nice one!

Alexander-Miller:
Pushed another small fix. Toggling behaviour should now work correctly in all cases.

PR #189, opened 2018-05-07
moving same treemacs function into treemacs-projectile

xdl:
Couldn't figure out why the fixes for the toggling issue (https://github.com/Alexander-Miller/treemacs/issues/183) didn't seem to be updated for my installation and then I remembered that I was using `treemacs-projectile` as well, updating it with the same code (apart from adding the `projectile-project-root` prompt) seems to fix it


Alexander-Miller:
Thanks :+1:

PR #209, opened 2018-06-08
Add fringe indicator mode

### LINKED ISSUE #172: Position indicator in the fringe
This one is for you @TimoFreiberg

Since some themes have a difficult to see hl-line-mode color treemacs should offer an additional visual helper, realized through a fringe bitmap.

Here's what needs to be done:
 * Define a face for the fringe bitmap
 * Create the bitmap itself
 * Create an overlay that will enable the display of the bitmap
 * Overlay must be buffer-local, and initialized as such
 * Create a minor mode to turn the feature on/off
 * Most of the code should go into treemacs-visuals.el
 * For the minor mode's setup & teardown the overlay probably needs to be destroyed & rebuilt.

TimoFreiberg:
Fixes #172


TimoFreiberg:
Looks like define-fringe-bitmap is not available in the emacs used by travis.
Maybe that emacs is built without a GUI toolkit (which can cause that error according to https://github.com/flycheck/flycheck/issues/57#issuecomment-12177590)

TimoFreiberg:
Another issue:
We need to think of another way to enable the minor modes by default, because setting the `init-value` to `t` doesn't run the setup-function. Maybe change the `init-value` to `nil` again and add `(treemacs-fringe-indicator-mode 1)` and the other 2 modes to `treemacs-mode-hook`?

Alexander-Miller:
Issue 1: we should have another sit-down on friday and have a look at 
elisp's compile time evaluation capabilities.

Issue 2: you can use the `only-during-treemacs-init` macro to run the 
enabling code. Just pack it into treemacs.el after all the require 
statements. I already do something like that for git-mode in 
treemacs-async.el, that part should be moved there as well.

TimoFreiberg:
Concerning issue 1:
How about
```
(setq treemacs--fringe-indicator-bitmap
      (if (boundp 'define-fringe-bitmap)
          (define-fringe-bitmap 'treemacs--fringe-indicator-bitmap
            (vector ...))
        'vertical-bar))
```
instead of
```
(define-fringe-bitmap 'treemacs--fringe-indicator-bitmap
            (vector ...))
```
The duplication of the symbol is ugly, but is the approach correct?

Alexander-Miller:
Well that's just practical and boring. 

I want to play with elisp's compile time code generation abilities anyway - I've this great idea for a debug-mode that's zero cost because the debug statement code isn't even generated unless you meet some predicate.

TimoFreiberg:
Yeah let's have some fun ;)
I already found a problem locally, `(boundp 'define-fringe-bitmap)` always returns nil...

Alexander-Miller:
In elisp each symbol has both a value and a function pointer. `boundp` checks the former, `fboundp` checks the latter ;)

TimoFreiberg:
stupid lisp-2s...

Alexander-Miller:
Should they get off your lawn?

TimoFreiberg:
I updated the MR with the boring version which works. (Incidentally, filewatch mode and follow mode don't seem to work on my machine, but let's look at that on friday)

Alexander-Miller:
Check if changing the init values is responsible for the breakage.

Compro-Prasad:
I think @TimoFreiberg 's git config isn't pointing to GitHub username.

src/elisp/treemacs-fringe-indicator.el
```diff
@@ -0,0 +1,100 @@
...
+;;; Commentary:
+;;; Handling of visuals in general and icons in particular.
+
+;;; Code:
+
+(require 'dash)
+(require 'treemacs-impl)
+(eval-and-compile (require 'treemacs-macros))
+
+(defcustom treemacs--fringe-indicator-bitmap
```
Alexander-Miller: Should go into customizations.el

src/elisp/treemacs-fringe-indicator.el
```diff
@@ -0,0 +1,100 @@
...
+      "The fringe bitmap used by the fringe-indicator minor mode."
+      :type 'fringe-bitmap
+      :options (if (fboundp 'fringe-bitmaps)
+                   (cons 'treemacs--fringe-indicator-bitmap fringe-bitmaps)
+                 '())
+      :group 'treemacs-configuration)
+
+(defvar-local treemacs--fringe-indicator-overlay nil)
+
+(defun treemacs--move-fringe-indicator-to-point ()
```
Alexander-Miller: Should be a defsubst.

src/elisp/treemacs-fringe-indicator.el
```diff
@@ -0,0 +1,100 @@
...
+                    #b00110000
+                    #b00110000
+                    #b00110000
+                    #b00110000))
+        'vertical-bar)
+      "The fringe bitmap used by the fringe-indicator minor mode."
+      :type 'fringe-bitmap
+      :options (if (fboundp 'fringe-bitmaps)
+                   (cons 'treemacs--fringe-indicator-bitmap fringe-bitmaps)
+                 '())
```
Alexander-Miller: Or nil?

Alexander-Miller:
@Compro-Prasad Private vs work laptop I guess.

PR #212, opened 2018-06-10
Fix hardcoded treemacs--persist-file

hlissner:
For those of us that'd like to centralize cache files and folders, this change:

1. Renames `treemacs--persist-file` to `treemacs-persist-file` (to imply that it is now public and safe to modify), 
2. Defines it with `defvar` so that, if a value has already been assigned to the variable, it won't be changed. `defconst` disregards customizations and resets its value much like `setq` does. 
3. Removes the hardcoded creation of the `~/.emacs.d/.cache` directory (`(f-mkdir user-emacs-directory ".cache")`.
4. Updates the reference to `treemacs-persist-file` in the README.

This makes it possible to truly customize where treemacs saves its persistence file.

EDIT: Or should I have used `defcustom`?

### DIFF
--- a/README.org
+++ b/README.org
@@ -209,8 +209,8 @@ Treemacs' sessions - your workspace and the projects it contains - are saved whe
 treemacs is first loaded. This persistence process is fully automatic and independant, and should therefore be fully
 compatible with ~desktop-save-mode~.
 
-The persisted state is saved under ~user-emacs-directory/.cache/treemacs-persist~. The exact file location
-is saved in the variable ~treemacs--persist-file~.
+The persisted state is saved under ~user-emacs-directory/.cache/treemacs-persist~ by default. The exact file location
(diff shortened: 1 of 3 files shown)


Alexander-Miller:
Yes defcustom is the way to go here - it's a customizable user option now. Specifically you'll need to

 * move the definition to treemacs-customization.el and require it here
 * add an entry in the config variable table in the readme
 * add it to the example installation snippet in the readme

Everything else looks good.

As a side note: while I stay on spacemacs nowadays I did try doom for a bit and what you did for neotree's visuals was just plain fantastic. If you ever do something similar for treemacs I hope you'll contribute your changes upstream. I've little in the way of artistic talent and greatly appreciate contributions in this area.

Alexander-Miller:
:+1:

Issue #204, opened 2018-06-01
Uneven height of treemacs window in mode-line


CSRaghunandan:
<img width="637" alt="screen shot 2018-06-01 at 5 52 36 pm" src="https://user-images.githubusercontent.com/5226809/40840519-adecc9d2-65c4-11e8-929f-0e1558794c1a.png">

Is this intentional by design?
How can I get the height to be the same as other buffers?

Alexander-Miller:
Not sure what's going on here. What modeline package do you use?

If you use spaceline then the treemacs modeline shows the current eyebrwose desktop number and then major-mode name, otherwise it's just a static text.

If you use something powerline-derived you can try chaning the value of `powerline-height`.

CSRaghunandan:
I'm using [moody](https://github.com/tarsius/moody/tree/adf652f35cba1bb3d0f254e1905e2deeeb0fbdba) package. Because of its simplicity compared to powerline or spaceline which required a lot of configuration to look decent.

Alexander-Miller:
Then I cannot say what's going on. Maybe you need to create a moodified mode-line-format for treemacs buffers or maybe some part of your usual modeline setup makes it larger than it would be if it only showed plain text.

CSRaghunandan:
Looks like I can't run `moody` mode-line for treemacs buffers. Is there a way I can set the height for treemacs mode-line format?

CSRaghunandan:
moody requires buffer identification ID. I get this error when I run moody in treemacs window: `Cannot find mode-line-buffer-identification and use moody-mode-line-buffer-identification in its place`

Can I add `mode-line-buffer-identification` for treemacs buffers?

Alexander-Miller:
If it's just a local value you can use set it to `buffer-name` in treemacs-mode-hook. All treemacs buffers have a unique name.

Alexander-Miller:
AFAIK there is no way to change the height of the modeline at all. I suggest you open an issue in the moody repo and ask how this value should be set.

CSRaghunandan:
Sure, I'll open an issue in Moody issues tracker

Alexander-Miller:
@CSRaghunandan  FYI: https://github.com/Alexander-Miller/treemacs/pull/214 is merged now.

PR #214, opened 2018-06-12
Add support for moody mode-line

dakra:
See https://github.com/Alexander-Miller/treemacs/issues/204 and the code is from @tarsius https://github.com/tarsius/moody/issues/17


Alexander-Miller:
:+1:

PR #197, opened 2018-05-12
Run ispell on docs.

perweij:
I ran ispell on README.org. It found some minor errors. It also proposed a few changes in casing - I can exclude the latter from the PR if you like.

PR #268, opened 2018-08-20
fix typo in README.org

bbigras:
This typo doesn't seem covered by #197


Alexander-Miller:
#197 is stuck. I'll first squash its commits myself and then integrate yours.

Alexander-Miller:
There we go.

PR #272, opened 2018-08-26
Preserve original file name when renaming it

br3athein:
IMO, most of the times it's better to have the original name as a default rename target, i.e. you'd like to only change a file extension or fix a tiny typo in a filename :slightly_smiling_face: 
`C-w` should do the trick if you don't like the original name one bit anyways


Alexander-Miller:
Thanks :+1:

PR #275, opened 2018-09-09
[README.org] Fix five 404 internal links on github

duianto:
problem:
Five internal links lead to 404 pages, for example these two:
```
[[Installation][below]]
[[Custom Icons][highly customizable]]
```

solution:
Use the same internal link names as in the TOC, with the leading pound symbol and the first letter in lowercase:
```
[[#installation][below]]
[[#custom-icons][highly customizable]]
```

Possible cause:
The links probably work fine in Emacs, because the Org manual: https://orgmode.org/manual/Internal-links.html
lists the current links as an option:
>Links such as ‘[[My Target]]’ or ‘[[My Target][Find my target]]’ lead to a text search in the current file.

but github doesn't seem to support text searching in org files, the links are instead translated like this:
https://github.com/Alexander-Miller/treemacs/blob/master/Installation
https://github.com/Alexander-Miller/treemacs/blob/master/Custom%20Icons

### DIFF
--- a/README.org
+++ b/README.org
@@ -95,7 +95,7 @@ If you don't care about reading the full readme here's a list of some bare bones
  * If you use projectile you can install ~treemacs-projectile~ to allow quickly add your projectile projects to
    treemacs.
  * Treemacs doesn't bind any global keys, you need to use whatever fits you best. A full install setup can be found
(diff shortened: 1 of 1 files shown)


Alexander-Miller:
:+1:  Thanks

PR #320, opened 2019-01-11
Add support for doom-modeline

BlindingDark:
Add support for [doom-modeline](https://github.com/seagle0128/doom-modeline)

before

after


Alexander-Miller:
:+1:

PR #354, opened 2019-02-12
Add error/warning/info icons

yyoncho:
I am not sure whether this is the correct way to define icons, so please advice.


Alexander-Miller:
The whole icon setup is more complicated that I'd like it to be what with terminal fallbacks, customization and resetability being all needed in the same spot. I don't have the whole thing in my head, but you can copy the way `treemacs-icon-root`, `treemacs-icon-root-text` and `treemacs-icon-root-png` are done. I'm not sure whether an empty string is the right fallback or using something like • would look better, but we can cross that bridge when we get there.

yyoncho:
Force pushed. FYI VSCode has 2 types of icons - one for light theme and one for dark theme. I have included the dark one.

src/elisp/treemacs-icons.el
```diff
@@ -134,6 +138,9 @@ Remember the value in `treemacs--default-icons-alist'."
 (defvar treemacs-icon-tag-leaf-text        (eval-when-compile (propertize "• " 'face 'font-lock-constant-face)))
 (defvar treemacs-icon-tag-node-closed-text (eval-when-compile (propertize "▸ " 'face 'font-lock-string-face)))
 (defvar treemacs-icon-tag-node-open-text   (eval-when-compile (propertize "▾ " 'face 'font-lock-string-face)))
+(defvar treemacs-icon-error-text           (eval-when-compile (propertize "• " 'face 'font-lock-string-face)))
```
Alexander-Miller: I wonder if this eval-when-compile even does anything. I gone a bit overboard with the whole compile-time-evaluation thing when I tried making loading of icons faster.

Alexander-Miller:
> one for light theme and one for dark theme

Interesting idea. Do you know if emacs has the means to decide whether some theme, or even just a color, is "dark" or "bright"?

yyoncho:
> > one for light theme and one for dark theme
> 
> Interesting idea. Do you know if emacs has the means to decide whether some theme, or even just a color, is "dark" or "bright"?

Nope, but I guess it is possible to convert the default bg colour to 3 hex numbers and compare them to 88 for example or something like that.

yyoncho:
I want to add icons for each object type (e. g. variable, class, method) if you mind. Is it possible to define a convenience method defining an icon?

yyoncho:
Also, one of the options is to use these - https://github.com/sebastiencs/company-box/tree/master/images or InteliJ one like https://pbs.twimg.com/media/D5K9Y4QWwAUrOo5.jpg:large

Alexander-Miller:
> Is it possible to define a convenience method defining an icon?

Not sure what you mean. `treemacs--setup-icon` can already be used without extensions, so all it does is define an icon variable.

> InteliJ one like https://pbs.twimg.com/media/D5K9Y4QWwAUrOo5.jpg:large

Looks like a case for themes. I think we have an issue for that somewhere. I'll write down my thoughts about how to do it over the weekend.

yyoncho:
> Not sure what you mean. `treemacs--setup-icon` can already be used without extensions, so all it does is define an icon variable.

What I am looking for is the ability to define everything in one place e. g. 

```
(treemacs-define-icon var-name image-path fallback-for-tui ...) 
```

Currently, in order to define a new icon you have to edit 6 places. (occurrences of `treemacs-icon-error`)

Alexander-Miller:
> Currently, in order to define a new icon you have to edit 6 places.

More macros it is then. I'll overhaul the whole icon thing again when I get to adding theme support.

Issue #337, opened 2019-02-05
Change default C-p Project prefix to C-c C-p.


rswgnu:
I see this topic has been broached before without success.  I know that I have not been following Treemacs evolution, so I am late on this issue but I will briefly try my hand at it.

Touch typists who use Emacs (and there are a lot of us) are definitely used to C-p moving to previous line most anywhere and they do not expect to have to rebind it to make another tool compatible with one of the most basic movement keys.  The fact that C-n works as expected makes this more of a default UI issue.

C-c C-p is a nice presently unused key sequence that would be a good alternative prefix for the lesser used Project commands and is only one key longer to type.  It jives with many other modes that have similar prefixes.

Although changing it represents a backward incompatibility, the benefit is that many more new users will be comfortable right away using the keyboard to move around in Treemacs.  Admitedly, n and p are inverse commands for movement by a line.

I know this is a hard issue for you but I thought I would give it a go.

rudi:
For what it's worth, this is the first thing I stumbled over as well, in an otherwise very nice package ;)

Here is what I did in my `.emacs` to adapt the keybindings to my taste; maybe the bug report can be left open so that others with the same problem can find this?

```elisp
(use-package treemacs :ensure t
  :config
  (unbind-key "C-p" treemacs-mode-map)
  (bind-key "C-n" 'treemacs-next-line treemacs-mode-map)
  (bind-key "C-p" 'treemacs-previous-line treemacs-mode-map)
  (bind-key "C-x p r" 'treemacs-rename-project treemacs-mode-map)
  (bind-key "C-x p a" 'treemacs-add-project-to-workspace treemacs-mode-map)
  (bind-key "C-x p d" 'treemacs-remove-project-from-workspace treemacs-mode-map)
  (bind-key "C-x p c c" 'treemacs-collapse-project treemacs-mode-map)
  (bind-key "C-x p c o" 'treemacs-collapse-other-projects treemacs-mode-map)
  (bind-key "C-x p c a" 'treemacs-collapse-all-projects treemacs-mode-map))
```

Note: if you use treemacs-projectile, the above needs to be in the `use-package` form for treemacs-projectile instead; you might want to add a binding for the `treemacs-projectile` command as well.

rudi:
[...]

PR #377, opened 2019-03-08
[Keybinds] Introduce keymaps for commands that share a common prefix

rudi:
- This makes it easy to adjust keybindings for commands in one place, even
  when they are defined in different source files.

- This change also makes it easier for users to adjust keybindings, as
  discussed in bug #337.

- Probably `treemacs-evil.el` could be modified accordingly as well; I have
  not investigated this further.

- This patch does not change any keybindings

### DIFF
--- a/src/elisp/treemacs-mode.el
+++ b/src/elisp/treemacs-mode.el
@@ -118,18 +118,18 @@ to it will instead show a blank."
               (format
                "
 %s
-%s              │ %s              │ %s    │ %s                │ %s              │ %s
+%s              │ %s              │ %s    │ %s                │ %s                │ %s
 ――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――
 %s next Line        │ %s dwim TAB            │ %s create file │ %s follow mode      │ %s add project    │ %s refresh
 %s prev line        │ %s dwim RET            │ %s create dir  │ %s filewatch mode   │ %s remove project │ %s (re)set width
 %s next neighbour   │ %s open no split       │ %s rename      │ %s git mode         │ %s rename project │ %s copy path
-%s prev neighbour   │ %s open horizontal     │ %s delete      │ %s show dotfiles    │                       │ %s copy root
-%s goto parent      │ %s open vertical       │ %s copy        │ %s resizability     │                       │ %s re-sort
-%s down next window │ %s open ace            │ %s move        │ %s fringe indicator │                       │ %s bookmark
-%s up next window   │ %s open ace horizontal │                    │                         │                       │
-                        │ %s open ace vertical   │                    │                         │                       │
-                        │ %s open externally     │                    │                         │                       │
-                        │ %s close parent        │                    │                         │                       │
+%s prev neighbour   │ %s open horizontal     │ %s delete      │ %s show dotfiles    │                         │ %s copy root
+%s goto parent      │ %s open vertical       │ %s copy        │ %s resizability     │                         │ %s re-sort
+%s down next window │ %s open ace            │ %s move        │ %s fringe indicator │                         │ %s bookmark
+%s up next window   │ %s open ace horizontal │                    │                         │                         │
+                        │ %s open ace vertical   │                    │                         │                         │
+                        │ %s open externally     │                    │                         │                         │
+                        │ %s close parent        │                    │                         │                         │
 "
                title
                column-nav               column-nodes          column-files           column-toggles          column-projects          column-misc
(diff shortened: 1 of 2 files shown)


rswgnu:
I like it.  Well done.

Alexander-Miller:
Evil keybinds still work, the helpful hydra still shows the right keys as well, so we're almost good to go. 

Just one last thing is missing: I have started to prefix my commit messages with an appropriate [keyword] for easier searching. For this PR that should be [Keybinds]. Also feel free to change the project prefix to `C-c C-p`, either here or in a new PR.

rudi:
Hi @Alexander-Miller, thanks for having a look!  I have adjusted the title of the pull request as requested, but it's not possible to change commit messages after the fact.  Should I submit a new pull request with adjusted commit message?  (Amending the commit will be too much of a hassle so I'd create a new branch.)

Alexander-Miller:
> it's not possible to change commit messages after the fact.

wat

> Amending the commit will be too much of a hassle

WAT

My first thought was that you're inexperienced with git, but you did use the technically correct term amend, and you did use a proper feature branch, signs that you know what you're doing. Those things are literally a `git commit --amend` and `git push --force` away, so you'll have to explain why you'd think that.

rudi:
Right you are.. force-pushing generally messes up everyone's repository states, but for a leaf commit I guess it's ok.  Pushed :)

rudi:
While I have your ear ... a long key sequence (i.e., `C-c C-p` as project prefix) leads to layout problems in the helpful hydra output (in some rows, the columns where that command is printed will be wider than the column heading) -- didn't try to fix this since it is only seen when the user explicitly reconfigures their treemacs.  Totally ok for me if you hold off on merging this pull request in case that's not acceptable for you.

Alexander-Miller:
> force-pushing generally messes up everyone's repository states,

Hence feature branches, doesn't matter how many times you force push what, so long as the end result is clean merging the PR is not a problem.

> a long key sequence (i.e., C-c C-p as project prefix) leads to layout problems in the helpful hydra output

The damage will be limited to minor adjustments if we change the prefix to `C-c p`, so let's do that, one bit of Ctrl less will still do what we want. Github actually lets you push on PR branches, so I just pushed that change myself, so we are ready to merge now :+1:

duianto:
There's a keybinding collision between `persp-mode` and Treemacs on `C-c p`. And the `persp-mode` bindings take precedence.

When `persp-mode` is enabled (default in Spacemacs), then the hydra looks like this:

emacs_2019-03-11_14-35-51

The `C-c p` bindings call:
- `a` `persp-add-buffer`
- `c` `persp-copy`
- `r` `persp-rename`

The only Treemacs project binding that still works is `C-c p d` for deleting a project.

rudi:
`C-c` followed by a lower-case or upper-case letter is reserved for the user -- see https://www.gnu.org/software/emacs/manual/html_node/elisp/Major-Mode-Conventions.html#Major-Mode-Conventions ("The key sequences bound in a major mode keymap should usually start with `C-c`, followed by a control character, a digit, or `{`, `}`, `<`, `>`, `:` or `;`. The other punctuation characters are reserved for minor modes, and ordinary letters are reserved for users.")

In summary -- neither treemacs nor persp-mode should bind C-c followed by an ordinary letter.

Note that evil / spacemacs likely have their own conventions, which I am unfamiliar with.

Note also that at least treemacs is a special mode (i.e., there's no ordinary text entry going on in the treemacs buffer), so has a lot of keys at its disposal that a normal mode wouldn't have.

rudi:
@duianto I apologize for the previous response -- on re-reading, I notice that you talked about a default configuration for spacemacs (which I am unfamiliar with), which makes this more of an issue.  But I think this pull request might be the wrong place to report this; it will be seen better if you file a separate bug in either of spacemacs / persp-mode / treemacs; hopefully they have guidelines for keybindings similar to what Emacs provides in the link I gave above.

Alexander-Miller:
See this is why I don't like changing keybinds. There's always something else that comes up.

Anyone opposed to the following plan: move project prefix to "P", rebind peek to "C-c P"?

duianto:
@rudi There's nothing to apologize for, you provided useful information, why `C-c` followed by a letter key shouldn't be used by packages, since they are reserved for users.

The reason for posting it here, was because this PR caused the issue by changing the Treemacs project keybindings.

The Treemacs projects prefix binding isn't a `persp-mode` issue. But your right that `persp-mode` shouldn't use a user reserved binding.

And it isn't a Spacemacs because:

The [Compability section of the Treemacs readme.org](https://github.com/Alexander-Miller/treemacs#compatibility) states that:
>Treemacs is included in Spacemacs and I am a Spacemacs user, therefore treemacs guarantees first-class support & compatibility for window-managing packages used in Spacemacs, namely persp, eyebrowse, popwin and window-purpose, as well as shackle.

The only keybindings that are defined in the Spacemacs Treemacs layer are the following:
(`spacemacs/set-leader-keys` is the prefix:
- `SPC` for the Evil and Hybrid editing styles in normal state.
- `M-m` for the Emacs editing style and for the Evil and Hybrid insert states.)
```
      (spacemacs/set-leader-keys
        "ft"    'treemacs
        "fB"    'treemacs-bookmark
        "fT"    'treemacs-find-file
        "f M-t" 'treemacs-find-tag
        "pt"    'spacemacs/treemacs-project-toggle)

```
source: https://github.com/syl20bnr/spacemacs/blob/4b195ddfc9228256361e0b264fe974aa86ed51a8/layers/%2Bfiletree/treemacs/packages.el#L48-L53

and the layer gives names to these Treemacs keybinding prefixes:
```
      (which-key-add-major-mode-key-based-replacements 'treemacs-mode
        "c"     "treemacs-create"
        "o"     "treemacs-visit-node"
        "oa"    "treemacs-visit-node-ace"
        "t"     "treemacs-toggles"
        "y"     "treemacs-copy"
        "C-p"   "treemacs-projects"
        "C-p c" "treemacs-projects-collapse"))
```
source: https://github.com/syl20bnr/spacemacs/blob/4b195ddfc9228256361e0b264fe974aa86ed51a8/layers/%2Bfiletree/treemacs/packages.el#L54-L61

The only thing that needs to be updated in Spacemacs when a Treemacs projects keybinding is chosen, is the which-key prefix name, and the documentation: https://github.com/syl20bnr/spacemacs/tree/develop/layers/%2Bfiletree/treemacs

duianto:
@Alexander-Miller In Evil mode maybe `p` could be used as the projects prefix, since it might not make sense to paste in the Treemacs window.

Currently `p` in evil mode is bound to the same command as `k`, `treemacs-previous-line`, but `n` might not be bound because it calls the default evil binding `evil-ex-search-next`.

[The treemacs documentation](https://github.com/Alexander-Miller/treemacs/blob/master/README.org#default-keymap) states that:

| Key           | Action                                      | Description                                                                                                    |
|---------------|---------------------------------------------|----------------------------------------------------------------------------------------------------------------|
| j/n           | treemacs-next-line                          | Go to the next line.                                                                                           |
| k/p           | treemacs-previous-line                      | Go to the previous line. 


I just learned about `P` (uppercase) for peek. That could probably stay in Evil mode as well.

I'm not sure what the Emacs bindings should be.

We should probably ask an Emacs bindings user like @rswgnu who opened the issue about changing the projects prefix:
Change default C-p Project prefix to C-c C-p. #337

Compro-Prasad:
How about switching to https://github.com/magit/transient ?

Alexander-Miller:
AFAIU transient is just a display library, it even used to use lv.el, same as hydra, so I don't see what good it'd do,.

Alexander-Miller:
> Anyone opposed to the following plan: 

Doesn't sound like it, so I'll be going through with it.

> The Compability section of the Treemacs readme.org states that:

That guarantee has recently been weakened since I've moved to spacemacs-base, so persp isn't even installed for me. I will need to be informed about such incompatibilities instead of running into them by default.

> In Evil mode maybe p could be used as the projects prefix, since it might not make sense to paste in the Treemacs window.

That just means having two problems instead of one. Unbinding p is fine, but the fewer package-specific keys and prefixes there are the better.

duianto:
>Anyone opposed to the following plan: move project prefix to "P", rebind peek to "C-c P"?

`C-c p` shouldn't be used since it's reserved for users.

In addition to the major mode conventions [mentioned by rudi above](https://github.com/Alexander-Miller/treemacs/pull/377#issuecomment-471548065)

The key binding conventions page states that:

>Don't define C-c letter as a key in Lisp programs. Sequences consisting of C-c and a letter (either upper or lower case) are reserved for users; they are the only sequences reserved for users, so do not block them.
Changing all the Emacs major modes to respect this convention was a lot of work; abandoning this convention would make that work go to waste, and inconvenience users. Please comply with it.

source: https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html

And I found this SO question:
https://stackoverflow.com/questions/31421106/why-emacs-project-c-c-p-is-undefined
where this comment https://stackoverflow.com/a/52046494 talks about `projectile` using the `C-c p` binding:
>This has changed a couple times in 2018. Boris used to set C-c p as the default leader, then changed it to C-c C-p to be in accordance with the emacs keybinding conventions. But now it's removed altogether, so you should set it yourself.

rswgnu:
> On Mar 12, 2019, at 2:07 AM, Alexander-Miller <notifications@github.com> wrote:
> 
> Anyone opposed to the following plan: move project prefix to "P", rebind peek to "C-c P"?
> 

I guess this is a little late but my thought is that since treemacs-project was on C-p, C-c C-p would be a good replacement in terms of muscle memory.  Is there any conflict with that?  P would  work if people can get used to hitting shift.

Bob

PR #391, opened 2019-03-23
[Extensions] Fixes to project extensions

Jasu:
This PR fixes two issues with project extensions:
- `treemacs-find-file-node` fails if it's looking for e.g. `README.md` and a project extension provides a node containing `README.md` before the file node. In that case, the regex search for `README.md` will match the extension node, and `treemacs-find-file-node` continues to treat the path as a string, and of course fails.
- `treemacs-find-node` and `treemacs-goto-node` don't handle project extension nodes at all. This causes an error e.g. when collapsing a project with an expanded project extension and then re-expanding it.

### DIFF
--- a/src/elisp/treemacs-impl.el
+++ b/src/elisp/treemacs-impl.el
@@ -778,62 +778,49 @@ failed."
             (goto-char start))
         search-result))))
 
-(define-inline treemacs--goto-custom-top-level-node (path)
-  (inline-letevals (path)
-    (inline-quote
-     (-when-let (result (treemacs--find-custom-top-level-node ,path))
-       (treemacs--evade-image)
-       (hl-line-highlight)
-       (set-window-point (get-buffer-window) (point))
-       result))))
-
-(defun treemacs--find-custom-dir-node (path)
-  "Move to the directory extension node at PATH."
-  (let* (;; go back here if the search fails
-         (project (treemacs--find-project-for-path (car path)))
-         (start (prog1 (point) (goto-char (treemacs-project->position project))))
-         ;; making a copy since the variable is a reference to a node actual path
-         ;; and will be changed in-place here
-         (goto-path (copy-sequence path))
-         (counter (1- (length goto-path)))
-         ;; manual as in to be expanded manually after we moved to the next closest node we can find
-         ;; in the dom
-         (manual-parts nil)
-         (dom-node nil))
-    ;; try to move as close as possible to the followed node, starting with its immediate parent
-    ;; keep moving upwards in the path we move to until reaching the root of the project (counter = 0)
-    ;; all the while collecting the parts of the path that beed manual expanding
-    (while (and (> (1+ counter) 0)
-                (null dom-node))
-      (setq dom-node (treemacs-find-in-dom (if (cdr goto-path) goto-path (car goto-path)))
-            counter (1- counter))
-      (cond
-       ((null dom-node)
-        (push (nth (1+ counter) goto-path) manual-parts)
-        (setcdr (nthcdr counter goto-path) nil))
-       ((and dom-node (null (treemacs-dom-node->position dom-node)))
-        (setq dom-node nil)
-        (push (nth (1+ counter) goto-path) manual-parts)
-        (setcdr (nthcdr counter goto-path) nil))))
-    (let* ((btn (if dom-node
-                    (treemacs-dom-node->position dom-node)
-                  (treemacs-project->position project)))
-           ;; do the rest manually
-           (search-result (if manual-parts (treemacs--follow-path-elements btn manual-parts) btn)))
-      (if (eq 'follow-failed search-result)
-          (prog1 nil
-            (goto-char start))
-        search-result))))
-
-(define-inline treemacs--goto-custom-dir-node (path)
-  "Move to the directory extension node at PATH."
-  (inline-letevals (path)
-    (inline-quote
-     (-when-let (result (treemacs--find-custom-dir-node ,path))
-       (treemacs--evade-image)
-       (hl-line-highlight)
-       (set-window-point (get-buffer-window) (point))
-       result))))
+(cl-macrolet
+    ((define-find-custom-node (name project-form doc)
+       `(defun ,name (path)
+          ,doc
+          (let* (;; go back here if the search fails
+                 (project (pop path))
+                 (start (prog1 (point) (goto-char (treemacs-project->position project))))
+                 ;; making a copy since the variable is a reference to a node actual path
+                 ;; and will be changed in-place here
+                 (goto-path (copy-sequence path))
+                 (counter (1- (length goto-path)))
+                 ;; manual as in to be expanded manually after we moved to the next closest node we can find
+                 ;; in the dom
+                 (manual-parts nil)
+                 (dom-node nil))
+            ;; try to move as close as possible to the followed node, starting with its immediate parent
+            ;; keep moving upwards in the path we move to until reaching the root of the project (counter = 0)
+            ;; all the while collecting the parts of the path that beed manual expanding
+            (while (and (> (1+ counter) 0)
+                        (null dom-node))
+              (setq dom-node (treemacs-find-in-dom (if (cdr goto-path) goto-path (car goto-path)))
+                    counter (1- counter))
+              (cond
+               ((null dom-node)
+                (push (nth (1+ counter) goto-path) manual-parts)
+                (setcdr (nthcdr counter goto-path) nil))
+               ((and dom-node (null (treemacs-dom-node->position dom-node)))
+                (setq dom-node nil)
+                (push (nth (1+ counter) goto-path) manual-parts)
+                (setcdr (nthcdr counter goto-path) nil))))
+            (let* ((btn (if dom-node
+                            (treemacs-dom-node->position dom-node)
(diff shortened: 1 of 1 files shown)


Alexander-Miller:
> if it's looking for e.g. README.md and a project extension provides a node containing README.md before the file node

As in two different nodes showing the same file? That is something treemacs assumes will never happen, at least for nodes with the exactly same `:path`. I am not sure about project extensions, though. Do you have a sample for the kind of extension you built to run into this? It's been a while since I touched that part of the code base and I would like to refamiliarise myself with it again to properly appreciate the changes you are making.

> treemacs-find-node and treemacs-goto-node don't handle project extension nodes at all.

Must be a remnant from when I decided that top-level extensions' paths should have a `:custom` prefix. The first identifying element used to be the pseudo-project instance of the extension, but that turned out to be a major hassle, especially for variadic top-level extensions.

You know the extension api was really the last module I expected to see PRs for. Not only is it the probably most complex part of the code base - you need to understand treemacs' architecture and how it's being generically used - it is also a barely field-test work in progress with outdated documentation (I intend to give the Extensions.org a work over once the api appears to look solid again). But then here you are on a roll, pumping out pull requests faster than I can review, so I am really curious what motivated this, is there something specific you are building with treemacs?

Alexander-Miller:
> In that case, the regex search for README.md

Nitpick: there's no regex involved, I use `search-forward` to search for the plain string.

src/elisp/treemacs-impl.el
```diff
@@ -886,10 +885,13 @@ PROJECT Project Struct"
     (treemacs-goto-file-node path project))
    ((eq :custom (car path))
     (treemacs--goto-custom-top-level-node path))
-   (t
-    (treemacs--goto-custom-dir-node path))))
+   ((stringp (car path))
+    (treemacs--goto-custom-dir-node path))
+   (t
```
Alexander-Miller: There's a trailing whitespace here. I recommend `(add-hook 'before-save-hook #'delete-trailing-whitespace)`.

src/elisp/treemacs-impl.el
```diff
@@ -886,10 +885,13 @@ PROJECT Project Struct"
...
     (treemacs--goto-custom-top-level-node path))
-   (t
-    (treemacs--goto-custom-dir-node path))))
+   ((stringp (car path))
+    (treemacs--goto-custom-dir-node path))
+   (t 
+    (treemacs--goto-custom-project-node path))))
 
 (defun treemacs-find-file-node (path &optional project)
+  (declare (debug t))
```
Alexander-Miller: You've put that debug above the eldoc. Looks like I need to look at what `make lint` is doing, this is the kind of stuff that should be caught by travis.

Alexander-Miller:
Something isn't quite right yet. With the Extensions.org example the Buffers node refuses to open whenever I refresh it when one of its buffer group nodes is expanded. I'll take a look at what this is about.

Jasu:
Thanks for the review, I'll take a look at the comments.

I'm working on an extension that shows the list of bookmarks at root, project, or directory level. I made the tab action to focus the file node in question in Treemacs. I uploaded it [here](https://github.com/Jasu/treemacs-bookmark) - it's obviously highly WIP and relies on the patches I sent.

The `treemacs-find-file-node` bug happens (for example) when I show bookmarks under a project, before directories (`'top`). I have a bookmark called `elpa`, and when I go to the file node, it searches the buffer for the filename (not path) `elpa` starting from the project root. Since the extension shows above the files, it finds the bookmark node first - only then it tries to compare paths, and that's where it fails.

> You know the extension api was really the last module I expected to see PRs for.

Are PRs still welcome though? I still have a few issues I'm willing to work on (but I have a busy week coming, so I'm not sure when). In particular, I still need to somehow solve the following issues:

- Refreshing the nodes when a bookmark is added
- Refreshing whether the extension root node should be visible at all (i.e. re-checking the predicate)
- Adding functions to open a file without a button with a Treemacs action - currently I have [this monstrosity](https://github.com/Jasu/treemacs-bookmark/blob/master/treemacs-bookmark.el#L91), and obviously I want to fix it before I put it anywhere near MELPA. I'd like to add functions to open a path (i.e. not depending on a button) to the public API, and make the actual Treemacs functions call those.

Alexander-Miller:
> Are PRs still welcome though?

Of course. Unexpected is very far from unwanted. I am quite glad someone else is also looking at and using the code. That usually helps weed out the bugs and edge cases I did not see myself.

I am off to bed soon and will continue the review tomorrow, but here's what I have so far: something is not right about the way you set up the goto-node code, in particular `treemacs--find-custom-project-node`. If you load the example extensions and run the following code you will see it finding the "Buffers" node instead of the "Buffer emacs-lisp-mode" node.

```Lisp
(with-current-buffer (treemacs-get-local-buffer)
  (treemacs--get-label-of
   (treemacs-goto-node
    (list (car (treemacs-workspace->projects (treemacs-current-workspace))) 'Buffers 'emacs-lisp-mode))))
```

Jasu:
> Of course. Unexpected is very far from unwanted. I am quite glad someone else is also looking at and using the code. That usually helps weed out the bugs and edge cases I did not see myself.

Awesome! I was afraid that I was going against the direction you want to take Treemacs to, or doing redundant work.

I'll try take a look at what managed to break before I go to sleep too, since I won't have time for that tomorrow.

src/elisp/treemacs-impl.el
```diff
@@ -773,14 +778,20 @@ failed."
...
-       result))))
+(defun treemacs--find-custom-project-node (path)
+  "Move to the project extension node at PATH."
+  (unless (treemacs-project->is-expanded? (car path))
+    (treemacs--expand-root-node (car path)))
+  (or (-some-> (treemacs-find-in-dom path)
+               (treemacs-dom-node->position))
+      (let* ((project (car path))
+             (start (point))
+             (project-start (treemacs-project->position project))
```
Alexander-Miller: If you need a project's start and end positions you can use `treemacs--get-bounds-of-project`.

src/elisp/treemacs-impl.el
```diff
@@ -773,14 +778,20 @@ failed."
...
+             (project-end (progn (goto-char project-start)
+                                 (treemacs--next-project-pos)))
+             (pos (text-property-any project-start project-end :key (nth 1 path))))
```
Jasu: And this is where the bug is. I somehow thought that the path consisted of project and key, not of project followed by path elements.

src/elisp/treemacs-impl.el
```diff
@@ -773,14 +778,20 @@ failed."
...
-       (treemacs--evade-image)
-       (hl-line-highlight)
-       (set-window-point (get-buffer-window) (point))
-       result))))
+(defun treemacs--find-custom-project-node (path)
+  "Move to the project extension node at PATH."
+  (unless (treemacs-project->is-expanded? (car path))
+    (treemacs--expand-root-node (car path)))
+  (or (-some-> (treemacs-find-in-dom path)
+               (treemacs-dom-node->position))
```
Jasu: I missed a `goto-char` here

src/elisp/treemacs-impl.el
Jasu: Also, I think there's a second bug here, as I think that searching for text properties requires the search key to be eq-comparable.

src/elisp/treemacs-impl.el
```diff
@@ -773,14 +778,20 @@ failed."
...
-    (inline-quote
-     (-when-let (result (treemacs--find-custom-top-level-node ,path))
-       (treemacs--evade-image)
-       (hl-line-highlight)
-       (set-window-point (get-buffer-window) (point))
-       result))))
+(defun treemacs--find-custom-project-node (path)
+  "Move to the project extension node at PATH."
+  (unless (treemacs-project->is-expanded? (car path))
+    (treemacs--expand-root-node (car path)))
```
Jasu: The whole path should be expanded here anyway

src/elisp/treemacs-impl.el
Jasu: Oops, my `.emacs.d` was missing `.editorconfig`. I hope the other PRs don't have trailing WS as well...

src/elisp/treemacs-impl.el
Jasu: I wasn't supposed to commit the declaratio at all, removed now.

src/elisp/treemacs-impl.el
Jasu: Thanks, I'll keep that in mind - I fixed it in an entirely different way, so that's no longer necessary here.

Jasu:
I pushed a new version - I reused the logic of `treemacs--find-custom-dir-node` with `cl-macrolet` instead of writing an entirely different traversal function. All current feedback should now be addressed.

Alexander-Miller:
Changes look fine now. Squash the fix commits and we're good to merge.

Jasu:
Nice! Should I squash the whole PR into a single commit, or do you want the two original commits as separate commits?

Alexander-Miller:
Let's go with the original 2 commits.

Jasu:
Ok, I rebased this on current master and force-pushed.

Alexander-Miller:
:+1:

PR #393, opened 2019-03-24
[Extensions] Manage projects correctly with extensions

Jasu:
**Problems**
- New projects are not inserted between top-level extensions
- Inserting a project clears the DOM if there are no other projects, but DOM might have extension-related data
- When there are no projects, separators are rendered incorrectly between top-level extensions
- When top-level extensions' predicate says no, whitespace might be rendered incorrectly
- When there are no projects, top-level extensions are not rendered.

**Solution**
 Keep a marker for the project add position, at the EOL of the last project. If no projects, at the EOL of the first extension positioned at 'top. When no extension, at BOB and not moving with insertion.
- When adding the first project, don't reset DOM because it might still contain extension-related data.
- Always render the projects, even when there are no projects, since  extensions might still be shown.
- Take into account the possiblity that nothing might follow or precede top-level expansions when adding the separator.
- Correctly handle top-level extension separators when the predicate  returns false.
- Extract separator insertion into an inlined function.
- When adding the first project, don't reset DOM because it might still contain extension-related data. **I'm not sure if there's a reason for clearing the DOM when adding the first project?**

### DIFF
--- a/src/elisp/treemacs-extensions.el
+++ b/src/elisp/treemacs-extensions.el
@@ -31,6 +31,8 @@
 
 (treemacs-import-functions-from "treemacs-mode"
   treemacs-mode)
+(treemacs-import-functions-from "treemacs-rendering"
+  treemacs--insert-root-separator)
 
 (defmacro treemacs--build-extension-addition (name)
   "Internal building block.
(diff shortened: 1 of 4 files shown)


Jasu:
This seems super complex due to all the corner cases. E.g opening and closing and re-opening a top-level extension will cause a trailing newline.

I think a lot of this complexity could be removed by adding automatically moving markers for the region where projects start and end and using those. This would also solve the issue that it's almost impossible to insert the first project between `'top` and `'bottom` top-level extensions.

Jasu:
I implemented this with a marker for the project insertion position.

src/elisp/treemacs-extensions.el
```diff
@@ -125,28 +127,33 @@ Also pass additional DATA to predicate function.")
...
+    treemacs--top-level-top-extensions
+    "Apply the top extensions for NODE of type `root' for the current WORKSPACE.
+
+Returns T if extensions were inserted.")
+  (define-root-extension-application
+    treemacs--apply-root-bottom-extensions
+    treemacs--top-level-bottom-extensions
+    "Apply the bottom extensions for NODE of type `root' for the current WORKSPACE.
+
+Returns T if extensions were inserted."))
```
Alexander-Miller: Pretty sure `t` isn't capitalized in elisp.

Jasu: I was probably just subconsciously trying to sneak some Common Lisp awesomeness into Emacs.

Fixed this one, and rebased on the current master.

Alexander-Miller:
:+1:

PR #394, opened 2019-03-24
[Workspaces] Fix workspace editing with top-level extensions

Jasu:
`treemacs--consolidate-projects` compares project paths as strings after finishing edit. If top-level extensions exist, this will fail.

This PR skips `:custom` buttons.


Alexander-Miller:
:+1:

PR #395, opened 2019-03-24
[Git] Fix git with open dir/project extensions

Jasu:
Don't try to run the git command with non-string paths.


Alexander-Miller:
:+1:

Issue #379, opened 2019-03-11
Spacemacs: Git mode causes 100 % CPU usage by treemacs-git-status.py


tsoernes:
Opening treemacs `v2.5-2019.03.03 @ 26.1` causes 100 % CPU churn on 2 cores. I've tried setting
```
  (setq-default treemacs-git-mode nil)
  (setq-default treemacs-use-git-mode nil)
  (setq-default treemacs-use-follow-mode nil)
  (fset #'trace-follow-hook #'ignore)
```
in `dotspacemacs/user-config` and also 
```
 '(treemacs-git-mode nil)
 '(treemacs-follow-mode nil)
```
in `dotspacemacs/emacs-custom-settings` but both git-mode and follow mode gets automatically enabled
causing 100 % churn by `treemacs-git-status.py`.

Alexander-Miller:
> (setq-default treemacs-git-mode nil)

> Setting this variable directly does not take effect;
> either customize it (see the info node `Easy Customization`)
> or call the function `treemacs-git-mode`.

> (fset #'trace-follow-hook #'ignore)

No such symbol exists, and even if it did as a hook it would've been a variable, not a function. This line does nothing.

> '(treemacs-git-mode nil)

Without context I have no idea what this does.

As for your actual issue take a look whether it's the same problem as in https://github.com/Alexander-Miller/treemacs/issues/369.

tsoernes:
@Alexander-Miller 
```
 '(treemacs-git-mode nil)
 '(treemacs-follow-mode nil)
```
in `custom-set-variables` in `dotspacemacs/emacs-custom-settings` are the result of using `customize-mode`. **These settings are not respected**. Despite their presence (and deleting the stuff in the user-config) after rebooting Emacs, both `git-mode` and `follow-mode` are enabled (customize-mode says their value is changed outside of customize, but there is no reference to treemacs at all in the dotfile outsize of the two lines that customize has put there.

This is not the same issue as #369. After killing `treemacs-git-status.py` manually, follow-mode is still enabled while git-mode is not and treemacs works without churning CPU. Treemacs chunks out because there's 150K files in the directory. Something is very inefficient there.

Alexander-Miller:
> in custom-set-variables in dotspacemacs/emacs-custom-settings are the result of using customize-mode. 

[...]

PR #403, opened 2019-04-08
[Git] Fix errors in find_recursive_entries

Jasu:
There was an argument missing in `find_recursive_entries`. 

Additionally, if the expanded directory contained a symlink loop, `treemacs-git-status.py` would hang indefinitely. Might <span>fix</span> #379 at least for some cases. This solution loses information about files behind symlinks though.


Alexander-Miller:
> There was an argument missing in find_recursive_entries

You'd think a language server would warn you about that kind of thing :/

:+1:

PR #405, opened 2019-04-09
[Interface] Skip empty lines in treemacs-next/previous-line

### LINKED ISSUE #401: Jump over empty lines between projects?
Could there be a use for having the cursor on an empty line between the projects?

If not, then maybe the cursor should jump over the empty lines and only move between the projects (or from a projects last dir/file to the next project).

Jasu:
Fixes #401.

* Use `forward-button` and `backward-button` to navigate between lines.
* Add a new customization option `treemacs-wrap-around`. When set, `treemacs-next-line` and `treemacs-previous-line` wrap around at the beginning and end of the buffer. The setting defaults to `nil`.

**Question:** Should the wrap setting default to `nil` or `t` in your opinion?


Jasu:
Rebased on `master`.

Alexander-Miller:
> Should the wrap setting default to nil or t in your opinion?

My philosophy is that new features should be enabled by default. There's already so many options and the readme is already so long that I doubt adding a new feature would ever be used by many people otherwise.

Jasu:
Made the option `t` by default and rebased on master.

Alexander-Miller:
:+1:

PR #407, opened 2019-04-12
[Extensions] Fix `treemacs--find-custom-top-level-node`

Jasu:
Top-level paths start with `(:custom Root-Key)`, i.e. the root path already contains two parts. Stop iterating earlier to account for that. Before, attempting to find a top-level extension's node that was not directly visible did not work. Also, finding the top-level extension root failed.


Alexander-Miller:
:+1:

PR #408, opened 2019-04-12
[Core] Fix treemacs-goto-node and treemacs-find-node

Jasu:
Non-existent string paths should return nil rather than executing default case of finding a custom project node.

Can be tested with `(treemacs-goto-node "something-that-does-not-exists")` and the same for `treemacs-find-node`. Previouly this raised an error.


Alexander-Miller:
:+1:

PR #409, opened 2019-04-12
[Compatibilty] Fix Emacs 25

Jasu:
Turns out that I broke Emacs 25. Emacs 25 does not have gensym by default, use make-symbol.


Alexander-Miller:
:+1:

PR #410, opened 2019-04-12
[Git] Fix handling of Git script warnings

### LINKED ISSUE #314: wrong-type-argument listp fatal:
Messages buffer:
```
Error running timer ‘treemacs--apply-deferred-git-state’: (wrong-type-argument listp fatal:) [2 times]
```

Jasu:
Currently, when `git status` gives a warning, e.g. `warning: could not open directory 'treemacs/.cask/25.3/elpa/gnupg/': Permission denied`, Treemacs `(read)`s this as `'warning:`. This causes a type error, making all files below the directory disappear from Treemacs. This pull

- Redirects git's `stderr` to `/dev/null` (using Python's `subprocess.DEVNULL`, so I'm at least guessing it will work on Windows), and
- Checks that the parsed output is a hash table.

**Edit:** I made this into two commits, in case you don't want the `stderr` one since it might make debugging some problems harder. If you want both commits, I can squeeze them to one if you want.

**Edit 2:** Fixes #314

**Edit 3:** Only including the `stderr` commit means that other warnings printed by Python could trigger the issue. Only including the parse result checking would not make the files disappear, but Git status would not be shown.

### DIFF
--- a/src/elisp/treemacs-async.el
+++ b/src/elisp/treemacs-async.el
@@ -155,7 +155,12 @@ GIT-FUTURE: Pfuture"
         (-let [git-output (pfuture-await-to-finish git-future)]
           (when (= 0 (process-exit-status git-future))
             (-let [parsed-output (read git-output)]
-              (when (hash-table-p parsed-output) parsed-output)))))
+              (if (hash-table-p parsed-output)
+                  parsed-output
+                (let ((inhibit-message t))
+                  (treemacs-log "treemacs-git-status.py output: %s" git-output))
+                (treemacs-log "treemacs-git-status.py did not output a valid hash table. See Messages buffer for details.")
+                nil)))))
       (ht)))
 
 (defun treemacs--git-status-process-simple (path)
--- a/src/elisp/treemacs-async.el
+++ b/src/elisp/treemacs-async.el
@@ -151,13 +151,12 @@ The real parsing and formatting is done by the python process. All that's really
 left to do is pick up the cons list and put it in a hash table.
 
 GIT-FUTURE: Pfuture"
-  (-let [ret (ht)]
-    (when git-future
-      (pfuture-await-to-finish git-future)
-      (when (= 0 (process-exit-status git-future))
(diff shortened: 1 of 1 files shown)


Alexander-Miller:
> Checks that the parsed output is a hash table.

Let's only include this part. Debugging other people's git problems is already quite difficult and hiding error output will only add another step to take.

Jasu:
Ok, force-pushed the commit away.

Jasu:
I added logging, since it just failed silently (on Emacs side). It prints two messages, one with the full output and one just telling that an error has occurred. The former message goes just to the `*Messages*` buffer, the latter gets displayed to the user.

Jasu:
This whole issue could have a better solution, but that would involve separating `stderr` from `stdout` in `pfuture`. I noticed that `pfuture` is also written by you, so would you be open for PRs there for this? I also noticed that it might have a race condition:

```elisp
...
         (process (apply #'start-process "Process Future" nil cmd cmd-args)))
    (process-put process 'result "")
    (set-process-filter process #'pfuture--append-output)
...
```

Since the process filter is not set when creating the process, some output might be lost if Emacs gets pre-empted between process creation and setting the process filter. If `make-process` was used, the process filter could be set immediately, and no buffer would have to be set for output, and a separate buffer could be set for `stderr`.

Alexander-Miller:
> I noticed that pfuture is also written by you, so would you be open for PRs there for this?

Sure. Cleaning up some of the interface and finishing its documentation has been a long time on my todo list now. melpa.org says that pfuture is only used by treemacs, so I am open even to breaking changes.

> If make-process was used, the process filter could be set immediately, and no buffer would have to be set for output,

That used to be the case, but IIRC I changed to the current version to be compatible with emacs 24. My MO is to support the current and the previous emacs version (or really whatever the people on stable distros have available), so currently we just need to be compatible with emacs 25.

Jasu:
I pushed a new solution to another branch, so I'll close this to keep the old simpler solution for reference.

PR #416, opened 2019-04-22
[WIP] An attempt to hide root button

Jasu:


src/elisp/treemacs-extensions.el
```diff
@@ -138,8 +138,7 @@ Also pass additional DATA to predicate function.")
                 (when (or (null predicate) (funcall predicate workspace))
                   (when (or has-previous (not is-first))
                     (treemacs--insert-root-separator))
-                  (setq is-first nil)
-                  (funcall extension))))
+                  (setq is-first (and is-first (funcall extension))))))
```
Jasu: Variadic extensions may print any number of root-level nodes, including zero. Each individual node will print the separator, so variadic extensions retunr `t` here.

src/elisp/treemacs-extensions.el
```diff
@@ -193,7 +192,8 @@ MORE-PROPERTIES is a plist of text properties that can arbitrarily added to the
 node for quick retrieval later."
   (treemacs-static-assert (and icon label-form state key-form)
     "All values except :more-properties and :face are mandatory")
-  `(list prefix ,icon
```
Jasu: If root-level nodes are printed, don't print prefix but print the project separator as a suffix instead.

src/elisp/treemacs-extensions.el
```diff
@@ -447,28 +449,33 @@ additional keys."
...
+                          (let ((pr (make-treemacs-project
                                      :name ,root-label
                                      :path ,root-key-form
-                                     :path-status 'extension)]
+                                     :path-status 'extension))
+                                (button-start (point)))
                             (treemacs--set-project-position ,root-key-form (point-marker))
                             (setq-local ,project-var-name pr)
-                            (insert (propertize "a"
+                            (insert (propertize "\n"
```
Jasu: Print a `\n`, since the button has to be on its own line. Otherwise, it will end up on the same line as some other button, and expanding it will insert the contents after the second button on the line.

src/elisp/treemacs-extensions.el
```diff
@@ -447,28 +449,33 @@ additional keys."
...
                                                 'button '(t)
                                                 'category 'default-button
+                                                'invisible t
+                                                'cursor-sensor-functions (list (lambda (_ previous-position change)
```
Jasu: `invisible` should not be selectable, but for some reason that's not the case. Use `cursor-sensor-mode` to skip to the next button on enter.

src/elisp/treemacs-extensions.el
```diff
@@ -447,28 +449,33 @@ additional keys."
...
+                                                                                 (when (eq change 'entered)
+                                                                                   (if (< previous-position (point))
+                                                                                       (treemacs-next-line 1)
+                                                                                     (treemacs-previous-line 1))
```
Jasu: Here are unhandled corner cases - if there are zero buttons, this will end up in an infinite loop. Same goes for wrapping around (or maybe it will just end up on the hidden button in that case)

src/elisp/treemacs-extensions.el
```diff
@@ -447,28 +449,33 @@ additional keys."
...
                                                 :project pr
                                                 :state ,closed-state-name))
-                            (goto-char 0)
-                            (funcall ',expand-name)
-                            (goto-char 0)
-                            ;; (narrow-to-region (1+ (point-at-eol)) (point-max))
-                            (setq-local hide-ov (make-overlay (point-at-bol) (1+ (point-at-eol))))
-                            (overlay-put hide-ov 'invisible t)
-                            ))))
+                            (save-excursion
```
Jasu: Not sure what's the correct behavior here. Probably not this... I believe the point should actually end up after the expanded nodes.

src/elisp/treemacs-mode.el
```diff
@@ -384,6 +384,8 @@ Will simply return `treemacs--eldoc-msg'."
   (when treemacs-fringe-indicator-mode
     (treemacs--setup-fringe-indicator-mode))
   (hl-line-mode t)
+  ;; Required for skipping hidden buttons
+  (cursor-sensor-mode t)
```
Jasu: `cursor-sensor-functions` requires this.

Jasu: This is supported in Emacs 25. `cursor-intangible-mode` is supposed to do precisely what we want, but for some reason or other it does not.

src/elisp/treemacs-extensions.el
```diff
@@ -401,7 +402,7 @@ additional keys."
          (treemacs--button-close
           :button btn
           :new-state ',closed-state-name
-          :new-icon ,(if icon-closed closed-icon-name icon-closed-form)
+          :new-icon ,(unless (equal top-level-marker (quote 'variadic)) (if icon-closed closed-icon-name icon-closed-form))
```
Alexander-Miller: Don't start cleaning this up. I have plans to take care of that constant variadic switching (basically another define-node-type method that will delegate to this one).

src/elisp/treemacs-mode.el
Alexander-Miller: That mode works by hooking into `pre-redisplay-functions`, so it gets called pretty often (at least 3 times just for inserting a character). Would it not be easier just to define our own post-command-hook?

src/elisp/treemacs-extensions.el
Jasu: Ok, I added that `unless` because the icon switching function tries to update the current button, and fails with the hidden node.

src/elisp/treemacs-mode.el
Jasu: Updated this to use post-command. I renamed the default-directory function to `treemacs--post-command` and put the logic there.

Jasu: ...And pushed again, to fix setting current dir and eldoc when avoiding invisible buttons.

Alexander-Miller:
Closing since your changes are included in https://github.com/Alexander-Miller/treemacs/commit/3897daa7ec1860e8f0555721a9cb9a1e2186198d.

PR #420, opened 2019-04-28
[Git] Fix handling of Git script warnings v2

Jasu:
* If pfuture-stderr is fbound, use it and show stderr to the user.
* If pfuture-result does not parse as a hash table, show a message to the user rather than causing an error.

With the current master of pfuture this discards git script output if it does not parse as a hash table.
With [this version](https://github.com/Alexander-Miller/pfuture/pull/5), it keeps working even when Git prints warnings to stderr. stderr is logged with messages.

### DIFF
--- a/src/elisp/treemacs-async.el
+++ b/src/elisp/treemacs-async.el
@@ -151,13 +151,33 @@ The real parsing and formatting is done by the python process. All that's really
 left to do is pick up the cons list and put it in a hash table.
 
 GIT-FUTURE: Pfuture"
-  (-let [ret (ht)]
-    (when git-future
-      (pfuture-await-to-finish git-future)
-      (when (= 0 (process-exit-status git-future))
-        (-let [git-output (pfuture-result git-future)]
-          (setf ret (read git-output)))))
-    ret))
+  (or (when git-future
+        (let* ((git-output (pfuture-await-to-finish git-future))
+               ;; Check fboundp in case an old version of pfuture is used.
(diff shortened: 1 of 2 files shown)


src/elisp/treemacs-async.el
```diff
@@ -151,13 +151,32 @@ The real parsing and formatting is done by the python process. All that's really
...
-  (-let [ret (ht)]
-    (when git-future
-      (pfuture-await-to-finish git-future)
-      (when (= 0 (process-exit-status git-future))
-        (-let [git-output (pfuture-result git-future)]
-          (setf ret (read git-output)))))
-    ret))
+  (or (when git-future
+        (let* ((git-output (pfuture-await-to-finish git-future))
+               ;; Check fboundp in case an old version of pfuture is used.
```
Alexander-Miller: I've now pushed a new tag for that change in pfuture, so you should also change the required pfuture version to 1.7.

Jasu: Should I still leave the `fboundp` check?

Alexander-Miller: Yes, let's leave it. People have little reason to update pfuture, so it'll be a while until the newest version proliferates (I do have a plan to write an update-treemacs function that pulls the latest treemacs & addons & pfuture versions). 

Best also leave a TODO comment to remove the check in about half a year's time. I usually go through these TODOs when releasing a new version.

Jasu:
I force-pushed the changes requested.

Alexander-Miller:
:+1:

Issue #421, opened 2019-04-29
[Git UI] folder shown as modified status if have gitignore file in it


BlindingDark:
![image](https://i.loli.net/2019/04/29/5cc6edde4a279.gif)

Alexander-Miller:
Investigating.

BlindingDark:
https://github.com/Alexander-Miller/treemacs/blob/7772b3e32fdd04d9c8623ddc4010b55edcf9cf44/src/scripts/treemacs-git-status.py#L67-L71

https://github.com/Alexander-Miller/treemacs/blob/7772b3e32fdd04d9c8623ddc4010b55edcf9cf44/src/scripts/treemacs-git-status.py#L75-L86

![image](https://user-images.githubusercontent.com/13600799/56939849-5ed97f00-6b3d-11e9-8cbd-c4218ce90680.png)

`!! cnfonts/v4/1~` not in `dirs_added` because it not end with `/`

BlindingDark:
I just noticed this commit.
https://github.com/Alexander-Miller/treemacs/commit/9f7ebad9e225ad091f7bd5eee26e5b7f63a3bf1c#diff-b5c7b399a46b13e6e27b5bedf3e394fcL57

Alexander-Miller:
Yep, I had removed that part because I could not remember anymore what it is good for. Thanks for the reminder.

PR #422, opened 2019-04-30
[Git] Fix gitignored file's directories as modified status.

BlindingDark:
https://github.com/Alexander-Miller/treemacs/issues/421


Alexander-Miller:
:+1:

PR #423, opened 2019-05-02
[Git] Don't lock Git index when updating status

Jasu:
Set GIT_OPTIONAL_LOCKS=0 in environment. Don't use --no-optional-locks, since it is only available in Git 2.14. This prevents errors with simultaneous git commands running alongside git status.

### DIFF
--- a/src/scripts/treemacs-git-status.py
+++ b/src/scripts/treemacs-git-status.py
@@ -1,5 +1,5 @@
 from subprocess import Popen, PIPE
-from os import listdir
+from os import listdir, environ
 from os.path import isdir, islink
 from posixpath import join
 import sys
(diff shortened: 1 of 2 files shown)


src/scripts/treemacs-git-status.py
```diff
@@ -36,6 +36,8 @@ def find_recursive_entries(path, state):
 
 def main():
     global output, ht_size
+    # Don't lock Git when updating status.
+    environ["GIT_OPTIONAL_LOCKS"] = "0"
```
Alexander-Miller: Do we really need both?

Jasu: The Emacs-side change was for the simple Git integration, while this one is for deferred.

Alexander-Miller: Ah, right.

Alexander-Miller:
:+1:

PR #424, opened 2019-05-02
[Core] Simplify treemacs-current-button

Jasu:
Make `treemacs-current-button` simpler and faster by using `text-property-not-all`.


Alexander-Miller:
:+1:

PR #429, opened 2019-05-07
[Core] Fix updating of custom root-level nodes

Jasu:
`treemacs--find-custom-top-level-node` must go to the node even if it is in DOM.

Otherwise, if a child node of a top-level extension is updated, `treemacs-do-update-node` tries to update the top-level node with the update tab action of its child.


Alexander-Miller:
:+1:

PR #433, opened 2019-05-09
Drag and drop support using left click(mouse 1)

### LINKED ISSUE #425: [Feature Request] Drag and drop a file to a specific window

Compro-Prasad:
Fixes #425

### DIFF
--- a/README.org
+++ b/README.org
@@ -177,6 +177,8 @@ A right click popup-menu is also available:
 
 [[file:screenshots/right-click.png]]
 
+You can also open a file in a specific window by dragging using left click from treemacs to the required window.
+
 ** Follow-mode
 ~treemacs-follow-mode~ is a global minor mode which allows the treemacs view to always move its focus to the currently
 selected file. This mode runs on an idle timer - the exact duration of inactivity (in seconds) before a move is called
--- a/src/elisp/treemacs-mouse-interface.el
+++ b/src/elisp/treemacs-mouse-interface.el
@@ -33,7 +33,7 @@
   "Move focus to the clicked line.
 Must be bound to a mouse click, or EVENT will not be supplied."
   (interactive "e")
-  (when (eq 'mouse-1 (elt event 0))
+  (when (eq 'down-mouse-1 (elt event 0))
     (select-window (->> event (cadr) (nth 0)))
     (goto-char (posn-point (cadr event)))
     (when (region-active-p)
@@ -94,6 +94,19 @@ Clicking on icons will expand a file's tags, just like
       (-when-let (state (treemacs--prop-at-point :state))
         (funcall (cdr (assoc state treemacs-doubleclick-actions-config)))))
     (treemacs--evade-image)))
+
+(defun treemacs-dragleftclick-action (event)
+  "Drag a file/dir node to be opened in a window.
+Must be bound to a mouse click, or EVENT will not be supplied."
+  (interactive "e")
+  (when (eq 'drag-mouse-1 (elt event 0))
+    (-when-let (treemacs-buffer (treemacs-get-local-buffer))
+      (let* ((node (with-current-buffer treemacs-buffer (treemacs-node-at-point)))
+             (path (-some-> node (treemacs-button-get :path))))
+        (treemacs-with-path path
+          :file-action (progn (select-window (elt (elt event 2) 0))
+                              (find-file path))
+          :no-match-action (ignore))))))
 
 (defun treemacs-define-doubleclick-action (state action)
   "Define the behaviour of `treemacs-doubleclick-action'.
--- a/src/extra/treemacs-evil.el
+++ b/src/extra/treemacs-evil.el
@@ -85,7 +85,8 @@ Then call ORIG-FUN with its ARGS and reopen treemacs if it was open before."
 (evil-define-key 'treemacs treemacs-mode-map (kbd "yy")     #'treemacs-copy-path-at-point)
 (evil-define-key 'treemacs treemacs-mode-map (kbd "yf")     #'treemacs-copy-file)
 (evil-define-key 'treemacs treemacs-mode-map (kbd "gr")     #'treemacs-refresh)
-(evil-define-key 'treemacs treemacs-mode-map [down-mouse-1] #'ignore)
+(evil-define-key 'treemacs treemacs-mode-map [down-mouse-1] #'treemacs-leftclick-action)
+(evil-define-key 'treemacs treemacs-mode-map [drag-mouse-1] #'treemacs-dragleftclick-action)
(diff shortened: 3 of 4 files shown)


src/elisp/treemacs-mouse-interface.el
```diff
@@ -29,6 +29,10 @@
 (require 'treemacs-filewatch-mode)
 (eval-and-compile (require 'treemacs-macros))
 
+(defvar treemacs-last-selected-button nil
```
Alexander-Miller: Use private double-dash.

src/elisp/treemacs-mouse-interface.el
```diff
@@ -48,6 +52,17 @@ Must be bound to a mouse click, or EVENT will not be supplied."
        :no-error            t))
     (treemacs--evade-image)))
 
+(defun treemacs-downleftclick-action (event)
```
Alexander-Miller: AFAICT this method is largely a duplication of the leftclick-action above, only used for saving the clicked node. If so we should combine them.

src/elisp/treemacs-mouse-interface.el
```diff
@@ -48,6 +52,17 @@ Must be bound to a mouse click, or EVENT will not be supplied."
...
+  "Select the window and record the node's path."
+  (interactive "e")
+  (when (eq 'down-mouse-1 (elt event 0))
+    (run-hooks 'mouse-leave-buffer-hook)
+    (select-window (elt (elt event 1) 0))
+    (goto-char (posn-point (cadr event)))
+    (setq treemacs-last-selected-button (treemacs-safe-button-get (treemacs-current-button) :path))
```
Alexander-Miller: The -safe-button-get is only needed when current-buffer and the button's buffer are different.

src/elisp/treemacs-mouse-interface.el
```diff
@@ -95,6 +110,12 @@ Clicking on icons will expand a file's tags, just like
         (funcall (cdr (assoc state treemacs-doubleclick-actions-config)))))
     (treemacs--evade-image)))
 
+(defun treemacs-dragleftclick-action (event)
+  (interactive "e")
+  (when (and (eq 'drag-mouse-1 (elt event 0)) treemacs-last-selected-button)
+    (select-window (elt (elt event 2) 0))
```
Alexander-Miller: The path is not checked for being a file.

src/extra/treemacs-evil.el
```diff
@@ -85,7 +85,8 @@ Then call ORIG-FUN with its ARGS and reopen treemacs if it was open before."
 (evil-define-key 'treemacs treemacs-mode-map (kbd "yy")     #'treemacs-copy-path-at-point)
 (evil-define-key 'treemacs treemacs-mode-map (kbd "yf")     #'treemacs-copy-file)
 (evil-define-key 'treemacs treemacs-mode-map (kbd "gr")     #'treemacs-refresh)
-(evil-define-key 'treemacs treemacs-mode-map [down-mouse-1] #'ignore)
```
Alexander-Miller: You should mention drag support in the readme under Detailed Feature List / Mouse Interface.

Alexander-Miller:
The commit message also needs the [Mouse] keyword.

src/elisp/treemacs-mouse-interface.el
Compro-Prasad: So, what should be used instead?

src/elisp/treemacs-mouse-interface.el
Compro-Prasad: Yeah. Will do that.

src/elisp/treemacs-mouse-interface.el
Alexander-Miller: Nothing, just the the -current-button as is.

Compro-Prasad:
I want to know how do you develop treemacs? I am currently having problem on how to select a specific git repository for developing a package. Spacemacs has this feature to use a git repo but for that I will have to change the actual code in spacemacs itself. Is there another way?

Alexander-Miller:
I don't go through spacemacs. Treemacs and its subsidiaries are all part of the excluded-packages list. Isntead I load treemacs straight from the hdd via use package. See https://github.com/Alexander-Miller/dotfiles/blob/master/.config/spacemacs/user-config.org#treemacs

src/elisp/treemacs-mouse-interface.el
Compro-Prasad: This has been resolved.

Compro-Prasad:
Is there something I am missing?

Alexander-Miller:
> Is there something I am missing?

Do you mean you're done with the requested changes? I think there should be a button somewhere for you to request a new review - that's what I was waiting for.

README.org
```diff
@@ -176,6 +176,8 @@ A right click popup-menu is also available:
 
 [[file:screenshots/right-click.png]]
 
+You can also open a file in a specific window by dragging using left click from ~treemacs~ to the required window.
```
Compro-Prasad: This is the added comment in the README.

Compro-Prasad:
I didn't know this feature of resolving conversations until you mentioned. I have resolved all o them from my side. Just test if this works.

README.org
```diff
@@ -14,45 +14,45 @@
 - [[#quick-feature-overview][Quick Feature Overview]]
 - [[#quick-start][Quick Start]]
 - [[#detailed-feature-list][Detailed Feature List]]
-   - [[#projects-and-workspaces][Projects and Workspaces]]
```
Alexander-Miller: Try regenerating the TOC with `org-list-indent-offset` set to 1. That *should* fix all those indent changes.

Compro-Prasad:
I have updated the TOC.

src/elisp/treemacs-macros.el
```diff
@@ -134,7 +134,7 @@ Will bind to current button to 'btn' for the executon of the action forms.
 When NO-ERROR is non-nil no error will be thrown if no match for the button
 state is achieved."
   (declare (debug (&rest [sexp form])))
-  `(-if-let (btn (treemacs-current-button))
+  `(-if-let (btn treemacs--last-selected-button)
```
Alexander-Miller: This change breaks treemacs, ensuring that TAB only applies to the last clicked button. This whole thing seems to be more complicated than it first appeared. I think it's best if I take over from here.

Compro-Prasad: Sure, you know the code better than me. You can take over.

Alexander-Miller: Pushed my changes. Please make sure it works on your end as well, and I'll merge.

Compro-Prasad: Just tested it. Works.

Alexander-Miller:
:+1:

PR #436, opened 2019-05-12
[Extensions] Add :visit-action to treemacs-define-*-node

Jasu:
This adds the final piece so I can get `treemacs-bookmark` ready. This lets users to define a global visit action, so that they don't have to be overridden separately for each extension. Extension API now has `:visit-action`, so extensions can use the window-splitting logic defined by Treemacs without having to reimplement all `RET` actions.

* Add a new customization option `treemacs-default-visit-action` and a node action `treemacs-visit-node-default`, to allow setting a global default visiting action.
* Make `treemacs-*-actions-configuration` variables use the default action instead of no-split.
* Make `treemacs-*-actions-configuration` variables use a custom widget for selecting the action.
* Add `:visit-action` to `treemacs-define-expandable-node` and `treemacs-define-leaf-node`. Its value is stored in the plist of the state symbol.
* Make `treemacs-execute-button-action` use the plist value of the state for unknow node types.
* Make `treemacs-define-*-node` default to `treemacs-visit-node-default` as their `:ret-action` if `:visit-action` is specified.
* Add missing documentation for `treemacs-doubleclick-actions-config.`


src/elisp/treemacs-interface.el
```diff
@@ -272,6 +272,11 @@ Stay in current window with a prefix argument ARG."
    :save-window arg
    :no-match-explanation "Node is neither a file, a directory or a tag - nothing to do here."))
 
+(defun treemacs-visit-node-default (&optional arg)
+  "Run `treemacs-default-visit-action' for the current button.
+A potential prefix ARG is passed on to the executed action, if possible."
+  (funcall treemacs-default-visit-action arg))
```
Alexander-Miller: Should be interactive, or defined elsewhere. The -interface module is for interactive, but not autoloaded functions.

Alexander-Miller:
> This adds the final piece so I can get treemacs-bookmark ready.

Take pictures when you do. A google image search for treemacs mostly brings up old bug reports.

src/elisp/treemacs-interface.el
Jasu: I'll make it interactive - I can imagine people wanting to bind this to some key.

Jasu:
Thanks for the review, I force-pushed the changes.

> Take pictures when you do. A google image search for treemacs mostly brings up old bug reports.

*In this picture, we see a typical Emacs coder finishing a Treemacs extension. Note the pride he takes in his work - he is wearing clean underwear, although every surface of his cave is covered in grease and Doritos crumbs.*

Alexander-Miller:
Nah, we are EMACS users, we have ORG-MODE. We like ORDER and STRUCTURE and NESTED LISTS. So many nested lists ...

:+1:

PR #437, opened 2019-05-12
Quote paths to commands that may contain spaces

nikclayton:
Previous code fails with:

version-to-list: Invalid version syntax: ‘'c:\Program' is not recognized as an internal or external command,
operable program or batch file.’ (must start with a number)

if the Python binary is in a path with a space, e.g. "c:\Program Files\...".

Quote the path with `shell-quote-argument` to fix this.


Alexander-Miller:
:+1:

Issue #331, opened 2019-01-28
Extension API feedback/questions


yyoncho:
Hi, I will use this PR for feedback/question on Extension API. Please let me know whether this is the preffered format for you.

1. How to define a function to return icon for treemacs-define-expandable-node like this:

```
(treemacs-define-expandable-node lsp-projects
  :icon-open (treemacs-icon-for-file (first item))
  :icon-closed (treemacs-icon-for-file (first item))
  :query-function (lsp-treemacs--get-files (treemacs-button-get btn :key))
  :render-action
  (treemacs-render-node ...)
```

2. How to define multiple top level roots when using treemacs to create standalone control? (the roots should be dynamic)

Alexander-Miller:
> How to define a function to return icon for treemacs-define-expandable-node like this

Ran into the same thing when working on buffers. Not possible now, the plan is to extend the api to pass either an `:icon-closed` that gets `defvar`d, or a dynamic `:icon-closed-form`. 

> How to define multiple top level roots when using treemacs to create standalone control? 

Here I'll need to think of something.

yyoncho:
> Here I'll need to think of something.

Skipping the root node and adjusting the indentation will do the job for me.

yyoncho:
3. `treemacs--evade-image` moves the cursor in the begining of the line even if you have navigated to different buffer as a result of ret action.

yyoncho:
4. What is the proper way to refresh standalone treemacs extension? I tried:

```
(with-current-buffer (get-buffer-create "*Showcase Buffer List*")
  (treemacs-update-node (list 'Buffers)))
```

It seems like the code is not adjusted for that case.

Alexander-Miller:
Dynamic icons have been pushed now. Variadic top-level extensions are next. I'll also investigate points 3 and 4.

Can you give me an example of what kind of data you'd use to initialize a variadic extension - is it just a list of strings or symbols or is there some additional data you want to make use of?

> Skipping the root node and adjusting the indentation will do the job for me.

You need to explain what you mean here. Which root node is being skipped? And how should indentation be adjusted? We're talking about top-level nodes so the indentation is just zero, right?

By the way, I made some changes to nomenclature to help avoid misunderstandings (if you use the old terms the compiler will warn you about their obsolescence). `treemacs-define-root-extension` is replaced by `treemacs-define-top-level-extension`, `:project-marker` is replaced by `:top-level-marker`. 

[...]

Issue #326, opened 2019-01-23
Buffer Display Extension


Alexander-Miller:
Let's dogfood my own extension api and build a display for buffers, similar to vsc. It's certainly been asked for often enough.
Some thoughts:

 * There's too many buffers in emacs, best group them by major mode (or something else).
 * Include customizable filters
 * I need *icons* for the root and group nodes (there's nothing good in vsc), single buffers' icons can be same as    their filetype mostly.
 * How to keep this thing up-to-date? There's a hook for killing buffers not nothing good for creation.
   I wouldn't want a hook running for every with-temp-buffer either. There are *a lot* of buffers thrown around.
 * See what ibuffer or other sidebars do.
 * What about follow-mode? hl-line-mode?

yyoncho:
IMO option to group them by project will be useful too .

As a side note, you may need to solve the multiple roots issue, right?

Alexander-Miller:
> As a side note, you may need to solve the multiple roots issue, right?

Not sure why you'd think that. Buffer nodes' paths will be quite different from the rest.

yyoncho:
Are you planning to nest buffers in the treemacs view or they will be in a separate window? If you have separate window IMO you will need multiple roots.

Alexander-Miller:
Same window. The buffers will be an extra "project" at the very top, same as vsc. Multiple windows to switch through is something to think about later.

Compro-Prasad:
I would suggest adding some keybindings to switch categorization and/or ordering of the buffers.
- Order:
  1. Most recent (time)
  2. Most used (count)
  3. Alphabetical
- Categories:
  1. No categories
  4. Project based
  5. Language based
  6. Major mode based
- Show/hide file and non file buffers
- Prioritize file and non file buffers

IMO adding so many keybindings will be hard but there should be a configuration option/function for the purpose which can be easily modified by the user.

Alexander-Miller:
> IMO option to group them by project will be useful too .
> I would suggest adding some keybindings to switch categorisation and/or ordering of the buffers.

First the main course, then the dessert. For now I'm stuck in implementation details (and the backlog).

> Are you planning to nest buffers in the treemacs view or they will be in a separate window?

Now that I think about it some kind of split window approach would have advantages. It would allow files and buffers to be visible at the same time and fix the follow/hl-line-mode question. Of course tracking and selecting those windows would be quite the nightmare in its own right.

> If you have separate window IMO you will need multiple roots.

[...]

PR #440, opened 2019-05-17
Make tag-follow-mode more robust in the presence of extensions

jaor:
When using (top level) extensions, treemacs can display nodes that are
neither tag nor file nodes.  And tag-follow-mode can end up saving one
of those nodes as its
treemacs--previously-followed-tag-position (imagine or instance a
buffer-follow-mode with an extension that lists buffers).  The code
modified here assumes that that cannot happen and the button at hand
must then be always a file node with a string :path, which leads to an
exception when treemacs-is-path uses string= on a path that is not a
string (because it's the path of, say, a buffer node).

I hope that makes sense.

The fix is trivial: just check first the kind of node, then its path.

I've also seen errors where the previous position wasn't valid, and
that's why i'm protecting the goto-char... i cannot reproduce them
reliably and they might well be due to other reasons, so maybe that
modification, altough harmless, is unnecessary.

### DIFF
--- a/src/elisp/treemacs-tag-follow-mode.el
+++ b/src/elisp/treemacs-tag-follow-mode.el
@@ -229,10 +229,10 @@ PROJECT: Project Struct"
                              (not (eq (car treemacs--previously-followed-tag-position) btn)))
                     (-let [(prev-followed-pos . prev-followed-path) treemacs--previously-followed-tag-position]
                       (save-excursion
-                        (goto-char prev-followed-pos)
-                        (when  (and (treemacs-is-path (-some-> (treemacs-current-button) (treemacs-button-get :path))
-                                                      :same-as prev-followed-path)
-                                    (eq 'file-node-open (treemacs-button-get prev-followed-pos :state)))
+                        (when (marker-position prev-followed-pos) (goto-char prev-followed-pos))
+                        (when  (and (eq 'file-node-open (treemacs-button-get prev-followed-pos :state))
+                                    (treemacs-is-path (-some-> (treemacs-current-button) (treemacs-button-get :path))
+                                                      :same-as prev-followed-path))
                           (treemacs--collapse-file-node prev-followed-pos)))))
                   ;; when that doesnt work move manually to the correct file
                   (-let [btn-path (treemacs-button-get btn :path)]


Alexander-Miller:
> And tag-follow-mode can end up saving one of those nodes as its treemacs--previously-followed-tag-position

Can or could? I don't see a situation where that would happen since `treemacs--previously-followed-tag-position` is only ever set by tag-follow-mode's internals, so it should always be a tag.

Alexander-Miller:
Besides, wouldn't it be simpler to just pull the `(eq 'file-node-open (treemacs-button-get prev-followed-pos :state))` condition to the front of the `and`? That would also ensure that we're on something we can close.

jaor:
On Sat, May 18 2019, Alexander-Miller wrote:

> Besides, wouldn't it be simpler to just pull the (eq 'file-node-open
> (treemacs-button-get prev-followed-pos :state)) condition to the front
> of the and? That would also ensure that we're on something we can
> close.

er, that's precisely what this patch does?

jaor:
On Sat, May 18 2019, Alexander-Miller wrote:

>     And tag-follow-mode can end up saving one of those nodes as its
>     treemacs--previously-followed-tag-position
>    
> Can or could? I don't see a situation where that would happen since
> treemacs--previously-followed-tag-position is only ever set by
> tag-follow-mode's internals, so it should always be a tag.

reading again the code of treemacs--do-follow-tag i see you must be
right on that. i've seen errors originated in treemacs--do-follow-tag
complaining of string= being called on a list rather than a string, and
my only explanation looking at that macro is that somehow
treemacs--previously-followed-tag-position got corrupted.  but i don't
have a good explanation for that.

jaor:
i cannot reproduce the issue anymore.  i guess i ended up in an inconsistent state while implementing my extension's follow-mode.  apologies for the noise, and thanks for your work on treemacs. it's a fun project to hack on.

Alexander-Miller:
> i've seen errors originated in treemacs--do-follow-tag complaining of
> string= being called on a list rather than a string

It'd be useful to know what the list was. There are node types, like
tags, whose :path property is a list, so knowing what node got we ran
into might give as a hint how it happened.

And since you're writing extensions as well you should be aware that
parts of the documentation in Extensions.org are out of date. The
extension api went through some changes due to #331, and re-adjusting
the docs is still on the agenda. The code of lsp-treemacs should provide
a good reference implementation.

Follow-mode integration isn't really part of the extension api, so
you're on your own for now, but you can expect that I'll think of
something as part of #326.

jaor:
On Sun, May 19 2019, Alexander-Miller wrote:

>> i've seen errors originated in treemacs--do-follow-tag complaining of
>> string= being called on a list rather than a string
>
> It'd be useful to know what the list was. There are node types, like
> tags, whose :path property is a list, so knowing what node got we ran
> into might give as a hint how it happened.

it was a "buffer path" (i was playing with an extension similar to the
buffers list extension in the docs).  if it happens again, i'll try to
zero in on the cause.

> And since you're writing extensions as well you should be aware that
> parts of the documentation in Extensions.org are out of date. The
> extension api went through some changes due to #331, and re-adjusting
> the docs is still on the agenda. The code of lsp-treemacs should provide
> a good reference implementation.

thanks, that's useful!

PR #443, opened 2019-05-21
[Core][Extensions] Make `find-*-node` go to the node

Jasu:
This fixes `treemacs-update-node` for project and dir nodes. Additionally, I made `find-file-node` go to the node, to harmonize the behavior of `find-*-node` functions, even if this does not fix any bug I'm aware of.


Alexander-Miller:
Let's stick with just the [Core] tag. I prefer to have just one main category. Other than that it LGTM.

Jasu:
Closing this, since I included the change in #444

PR #449, opened 2019-05-29
[Compatibility] Improve python3 detection on windows

roadrunner1776:
Bug fix: don't fail on (Windows) systems without python installed

### DIFF
--- a/src/elisp/treemacs-customization.el
+++ b/src/elisp/treemacs-customization.el
@@ -32,11 +32,12 @@
            (s-trim)
            (s-lines)
            (--first
-            (->> (concat (shell-quote-argument it) " --version")
-                 (shell-command-to-string)
-                 (s-trim)
-                 (s-replace "Python " "")
-                 (version<= "3")))))))
+	    (when (not (file-exists-p it))
+	      (->> (concat (shell-quote-argument it) " --version")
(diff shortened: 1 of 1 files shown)


src/elisp/treemacs-customization.el
```diff
@@ -32,11 +32,12 @@
            (s-trim)
            (s-lines)
            (--first
-            (->> (concat (shell-quote-argument it) " --version")
+            (if (file-exists-p it)
```
Alexander-Miller: Without an else branch this should be a `ẁhen`.

roadrunner1776: Replaced the if statement with a when statement

Alexander-Miller:
:+1:

duianto:
This current version returns `nil` for me:
https://github.com/Alexander-Miller/treemacs/blob/ba34423a12bcdbf74e0afa4bd4bd5ff877251c08/src/elisp/treemacs-customization.el#L26

The previous version returns the path to python 3.
`"c:\\Python37\\python.exe"`

https://github.com/Alexander-Miller/treemacs/blob/2b652d8273de012bc3a825cfc6cc73f4c03906da/src/elisp/treemacs-customization.el#L25

Issue #453, opened 2019-06-03
Issue with git-mode deferred and extended


mpanarin:
Hello
With Emacs 27 and spacemacs I get this error, when trying to use git-mode `deferred` or `extended`:
```
Treemacs-Git mode enabled
[Treemacs] Refresh complete.
[Treemacs] treemacs-git-status.py wrote to stderr (see full output in *Messages*): Traceback (most recent call last):,   File "/home/m-panarin/.emacs.d/elpa/27....
[Treemacs] treemacs-git-status.py wrote to stderr: Traceback (most recent call last):
  File "/home/m-panarin/.emacs.d/elpa/27.0/develop/treemacs-20190601.2254/treemacs-git-status.py", line 1, in <module>
    from subprocess import Popen, PIPE
ModuleNotFoundError: No module named 'subprocess'
```

clearly it is a python issue, but that is a built-in module and i can import it without issues in my interpreter, as well as I can run this script without issues with my python3 interpreter, so I am not sure where to dig.
`simple` mode works fine.

mpanarin:
issue disappeared by itself. Don't know what was wrong :\

PR #456, opened 2019-06-12
[FIX] remove `-S` flag from python status call

mpanarin:
Connected to the issue https://github.com/Alexander-Miller/treemacs/issues/453

I noticed that I have this issue only inside virtualenvironments created with `virtualenvwrapper`, this fixes the issue.


Alexander-Miller:
You should remove the other instances of its use as well. It's all in the same file.

Even if that is not strictly necessary, I'd prefer things be consistent,

Alexander-Miller:
Oh and let's start the commit message with `[Misc]`.

mpanarin:
@Alexander-Miller done

Alexander-Miller:
:+1:

PR #458, opened 2019-06-13
[Icons] Add icons for TeX and Racket

emacsomancer:
- Racket icon from
  https://en.wikipedia.org/wiki/Racket_(programming_language)#/media/File:Racket-logo.svg

- TeX icon generated in TeX via `\color{red}\bfseries\TeX`


Alexander-Miller:
:+1:

Issue #400, opened 2019-04-03
Update readme project management bindings


duianto:
The project management bindings were discussed and decided in this PR:
https://github.com/Alexander-Miller/treemacs/pull/377#issuecomment-472538569

The projects prefix bindings are now: `C-c C-p`

But the readme still shows the prefix as: `C-p`

The last line in this section:
https://github.com/Alexander-Miller/treemacs#quick-start

The first line in this section:
https://github.com/Alexander-Miller/treemacs#conveniently-editing-your-projects-and-workspaces

And in the Default keymap section:
https://github.com/Alexander-Miller/treemacs#default-keymap

PR #461, opened 2019-06-19
[Docs] Update projects prefix key bindings

duianto:
Fixes: #400

### DIFF
--- a/README.org
+++ b/README.org
@@ -106,7 +106,7 @@ If you don't care about reading the full readme here's a list of some bare bones
    move between projects.
  * There's half a dozen different ways to open nodes, all bound under o as prefix. Pick your favourite.
  * TAB and RET are particularly configurable. See ~treemacs-TAB/RET-actions-config~.
- * Projects administration is bound under the ~C-p~ prefix.
+ * Projects administration is bound under the ~C-c C-p~ prefix.
 
 * Detailed Feature List
 ** Projects and Workspaces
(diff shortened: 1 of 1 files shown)


Compro-Prasad:
Emacs doesn't recommend using the `C-c` prefix in packages. It is meant only for user space keybindings. So, some different keybinding should be used. `C-x p` is not bound to any command so we can use it instead. Or `C-x C-p` too, which is `mark-page` which I don't think is used by many people.

duianto:
This was discussed in this issue:
[Keybinds] Introduce keymaps for commands that share a common prefix #377

The convention seems to be that `C-c` followed by an upper or lower case letter are the only sequence reserved for users.

>Don't define C-c letter as a key in Lisp programs. Sequences consisting of C-c and a letter (either upper or lower case) are reserved for users; they are the only sequences reserved for users, so do not block them.
Changing all the Emacs major modes to respect this convention was a lot of work; abandoning this convention would make that work go to waste, and inconvenience users. Please comply with it.

source: https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html

>The key sequences bound in a major mode keymap should usually start with C-c, followed by a control character, a digit, or {, }, <, >, : or ;. The other punctuation characters are reserved for minor modes, and ordinary letters are reserved for users.

source: https://www.gnu.org/software/emacs/manual/html_node/elisp/Major-Mode-Conventions.html#Major-Mode-Conventions

Alexander-Miller:
Thanks for the fix. The commit message should start with `[Docs]` - I sort my commits by category rather than files.

Any while we're at it, you wouldn't happen to know anything about that:

> /tmp/emacs/bin/emacs: error while loading shared libraries: libgnutls.so.30: cannot open shared object file: No such file or directory

I haven't touched my build in ages, and now it's all failing due to this.

duianto:
I'm not familiar with automatic testing.

When I searched for the error message, I found this page:
http://emacs.1067599.n8.nabble.com/emacs-update-on-archlinux-td356260.html
You might already have found/read it. Otherwise maybe there's some useful info there.

duianto:
I asked JAremko about the error, and got this response:
>mb previously his emacs version had libgnutls linked statically or something(mb it didn't use it at all). And now it's dynamic libgnutls that must be installed on OS
Installing gnutls package should solve this.

Alexander-Miller:
The build fix was easier than expected. I guess travis changed something about their VM setup, and changing the language declaration from `emacs-lisp` to `generic` fixed things.

Issue #435, opened 2019-05-11
org-mode headings face shown incorrectly


llcc:
I installed the recent version, and bugs appeared as shown below:

![image](https://user-images.githubusercontent.com/3996879/57568795-136d7d80-7427-11e9-86cb-1d919010ff01.png)

Thanks!

Alexander-Miller:
Do the colors change correctly when you load another theme?

llcc:
@Alexander-Miller I switched to other themes、but the problem remains.

Alexander-Miller:
What are the values of
 - `treemacs--selected-icon-background`
 - `treemacs--not-selected-icon-background`
 - `treemacs--created-icons`

llcc:
- `treemacs--selected-icon-background`  gives `"#eee8d5"`
- `treemacs--not-selected-icon-background` gives `"#fdf6e3"`
- `treemacs--created-icons` gives
  ```
  (#("  " 0 1 (img-unselected (image :type png :file ".emacs.d/elpa/treemacs-20190507.1902/icons/vsc/reason.png" :scale 1.203968253968254 :ascent center :background "#fdf6e3") img-selected (image :type png :file ".emacs.d/elpa/treemacs-20190507.1902/icons/vsc/reason.png" :scale 1.203968253968254 :ascent center :background "#eee8d5") display (image :type png :file ".emacs.d/elpa/treemacs-20190507.1902/icons/vsc/reason.png" :scale 1.203968253968254 :ascent center :background "#fdf6e3")))
  #("  " 0 1 (img-unselected (image :type png :file ".emacs.d/elpa/treemacs-20190507.1902/icons/vsc/r.png" :scale 1.203968253968254 :ascent center :background "#fdf6e3") img-selected (image :type png :file ".emacs.d/elpa/treemacs-20190507.1902/icons/vsc/r.png" :scale 1.203968253968254 :ascent center :background "#eee8d5") display (image :type png :file ".emacs.d/elpa/treemacs-20190507.1902/icons/vsc/r.png" :scale 1.203968253968254 :ascent center :background "#fdf6e3")))
[...]

PR #469, opened 2019-06-27
[Icon] Fix png background.

BlindingDark:
#435

I re-converted the format with GIMP and it looks normal to me.
Seems that the original picture lost some background color information.

---

Waiting for @llcc to confirm if the problem is also solved.


Alexander-Miller:
I pulled those icons as .xpms from emacs itself and converted with imagemagick. I guess that's when the background got broken.

:+1:

llcc:
Hi @Alexander-Miller, sorry to reply late. 
@BlindingDark solved the problem.

PR #471, opened 2019-06-27
Add trailing '/' so that `f-move` doesn't throw any error

Compro-Prasad:
This bug occurs when read-directory-name opens a GUI dialog box and
asks for a directory. This way the selected directory doesn't end with
a trailing '/'. This commit fixes that.

Reproduction steps:

1. Open Right Click menu
2. Click on *Move*

This will bring up the GUI dialog box. Choose a destination directory
and confirm. This will throw an error without this commit.


Compro-Prasad:
What tag should I add to this commit @Alexander-Miller ?

Alexander-Miller:
`[File Management]` is a good fit.

Compro-Prasad:
Done

Alexander-Miller:
:+1:

Issue #473, opened 2019-06-28
Error when launching treemacs after update: "Invalid version syntax"


sandoichi:
I'm using spacemacs and recently updated all of my packages. I now have `treemacs-20190628.532` and when I launch it, I get the following error message 
`version-to-list: Invalid version syntax: ‘The filename, directory name, or volume label syntax is incorrect.’ (must start with a number)` and it never actually loads. Not really sure how to debug further.

I'm using emacs 26.1 and the latest spacemacs develop branch

Alexander-Miller:
Installing the layer from melpa works on my end. 
Please provide a full stack trace after running `toggle-debug-on-error`.

sandoichi:
`Debugger entered--Lisp error: (error "Invalid version syntax: ‘The filename, directory name, or volume label syntax is incorrect.’ (must start with a number)")
  signal(error ("Invalid version syntax: ‘The filename, directory name, or volume label syntax is incorrect.’ (must start with a number)"))
  error("Invalid version syntax: `%s' (must start with a number)" "The filename, directory name, or volume label syntax is incorrect.")
  version-to-list("The filename, directory name, or volume label syntax is incorrect.")
  version<=("3" "The filename, directory name, or volume label syntax is incorrect.")
  treemacs--find-python3()
  #f(compiled-function () #<bytecode 0x4dd6759>)()
  funcall(#f(compiled-function () #<bytecode 0x4dd6759>))
  eval((funcall (function #f(compiled-function () #<bytecode 0x4dd6759>))))
  custom-initialize-reset(treemacs-python-executable (funcall (function #f(compiled-function () #<bytecode 0x4dd6759>))))
  custom-declare-variable(treemacs-python-executable (funcall (function #f(compiled-function () #<bytecode 0x4dd6759>))) "The python executable used by treemacs.\nAn asynchronous python process is used in two optional feaures:\n`treemacs-collapse-dirs' and the extended variant of `treemacs-git-mode'.\n\nThere is generally only one reason to change this value: an extended\n`treemacs-git-mode' requires python3 to work. If the default python executable\nis pointing to python2 this config variable can be used to direct treemacs to\nthe python3 binary." :type string :group treemacs-git)
[...]

PR #481, opened 2019-07-08
[Core] Use find python 3 command with windows support

duianto:
problem:
(executable-find "python3") doesn't work in windows, because the executable is
called python.exe, and it's located in a python37 directory.

solution:
use (treemacs--find-python3) which has windows support


Alexander-Miller:
:+1:

And while you're already here can you take a look at https://github.com/Alexander-Miller/treemacs/issues/473? I'd like to make sure that finding python on windows behave as I assume it does.

PR #509, opened 2019-08-06
[Extensions] Fix variadic extensions positioned at top

Jasu:
This fixes the positioning of variadic extensions at top by removing a `save-excursion`. The `save-excursion` made the point be before the variadic root-level nodes, causing projects to always be inserted before the extension nodes.

PR #511, opened 2019-08-09
[Extensions] Fix variadic extensions at top, version 2

Jasu:
This is a rather large change.

- Buttons now include the whitespace and icon at their beginning. This makes it possible to add custom keymaps to the buttons. I have a branch for that, to add custom create/copy/rename/etc. actions for extension buttons. There was also some reason to do that here, but I forget what it was. Maybe it even isn't necessary anymore after all the iterations I made to this.
- Buttons "own" any trailing newlines.
- Trailing newlines are printed unconditionally.
- Motion is constrained to the last project in `treemacs--post-command`

I printed the `buffer-string` after various operations involving variadic extensions, and could not break them this time. The hidden nodes always seem to be on their own lines.

I could not manage to narrow the buffer to the visible part even this way, since that could have meant narrowing empty variadic extension buttons away.

I also didn't manage to make the project separator use a display property for inter-project space, since `hl-line-mode` would paint the whole two-line separator when a project would be hovered.

### DIFF
--- a/src/elisp/treemacs-core-utils.el
+++ b/src/elisp/treemacs-core-utils.el
@@ -340,11 +340,10 @@ Returns nil if no such buffer exists.."
   "Replace icon in current line with NEW-SYM."
   (inline-letevals (new-sym)
     (inline-quote
-     (save-excursion
-       (let ((len (length ,new-sym)))
-         (goto-char (- (treemacs-button-start (next-button (point-at-bol) t)) len))
-         (insert ,new-sym)
-         (delete-char len))))))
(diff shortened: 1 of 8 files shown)


Alexander-Miller:
Closing for house keeping.

Alexander-Miller:
Wait, I thought this was the continued in #509, not the other way around.

It's been a while and I lost track of things. Do you want to get this merged, or would you rather realize the plans you mentioned in #509? (And I don't mean do you want to deliver the full working PR, just putting things down in an issue will be well enough)

PR #520, opened 2019-08-24
[Icons] Add dhall icon

aveltras:


Alexander-Miller:
Are there alternative versions? A black icon doesn't mix well with dark themes.

aveltras:
I used it with a dark theme, there is a thin white border around it.
I guess this might be better to apply another color on it like i did for the purescript logo.
I'll come back to this later.

PR #530, opened 2019-09-12
[Compatibility] switch to regexes in winum ignore

mpanarin:
As https://github.com/deb0ch/emacs-winum/pull/22 was merged, we can switch to regex now

### DIFF
--- a/src/elisp/treemacs-compatibility.el
+++ b/src/elisp/treemacs-compatibility.el
@@ -33,10 +33,8 @@
  (push '(treemacs-workspace . :never) frameset-filter-alist))
 
 (with-eval-after-load 'winum
-  (when (boundp 'winum-ignored-buffers)
-    (dolist (n (number-sequence 1 5))
-      (add-to-list 'winum-ignored-buffers
-                   (format "%sFramebuffer-%s*" treemacs--buffer-name-prefix n)))))
+  (when (boundp 'winum-ignored-buffers-regexp)
+    (add-to-list 'winum-ignored-buffers-regexp (regexp-quote (format "%sFramebuffer-" treemacs--buffer-name-prefix)))))
 
 (with-eval-after-load 'ace-window
   (when (boundp 'aw-ignored-buffers)


src/elisp/treemacs-compatibility.el
```diff
@@ -33,10 +33,8 @@
  (push '(treemacs-workspace . :never) frameset-filter-alist))
 
 (with-eval-after-load 'winum
-  (when (boundp 'winum-ignored-buffers)
-    (dolist (n (number-sequence 1 5))
-      (add-to-list 'winum-ignored-buffers
-                   (format "%sFramebuffer-%s*" treemacs--buffer-name-prefix n)))))
+  (when (boundp 'winum-ignored-buffers-regexp)
+    (add-to-list 'winum-ignored-buffers-regexp (format "%sFramebuffer-" treemacs--buffer-name-prefix))))
```
Alexander-Miller: The regex isn't right and only works by chance. There's a `*` in the prefix so it matches whatever and I don't see anything to match against the buffer number at the end. This one should work:

> `(rx-to-string `(seq ,treemacs--buffer-name-prefix "Framebuffer-" (1+ num)))`

mpanarin: Well, that is just not true.
```elisp
(string-match " *treemacs-Framebuffer-" "*treemacs-Framebuffer-1")  ;; evaluates to 1
(string-match " *treemacs-Framebuffer-" "*treemacs-Framebuffer-2")  ;; evaluates to 1
(string-match " *treemacs-Framebuffer-" "*tremacs-Framebuffer-1")  ;; evaluates to nil
```

About the number, shouldn't any treemacs framebuffer be ignored by winum?

Alexander-Miller: Ah, you're right about the number, I was expecting  to match the full string, but leaving it out should work as well.  Still, the regex should still go through `regexp-quote` to make sure that the `*` is quoted properly.

mpanarin: updated

Alexander-Miller:
:+1:

PR #544, opened 2019-10-07
[Icons] Let .gitattributes extension to show git icon

Asalle:


Asalle:
I don't understand, does ci fail because some package is unavailable? Did i do something?

Alexander-Miller:
:+1: 

> Did i do something?

Nope, CI has been a doofus lately, and is always failing to install evil for whatever reason while it's working just fine locally.

PR #546, opened 2019-10-08
[Rendering] Fix create ingnored file.

BlindingDark:
I found some strange renamed events, such as `old-name` is "recentf" and `new-name` is "recentf~".

https://github.com/Alexander-Miller/treemacs/blob/d653a14ae8aea89d166edb825580fde31c8328f1/src/elisp/treemacs-filewatch-mode.el#L178-L185

This will cause ignored file to be displayed.

image

### DIFF
--- a/src/elisp/treemacs-filewatch-mode.el
+++ b/src/elisp/treemacs-filewatch-mode.el
@@ -181,7 +181,8 @@ file from caches if it has been deleted instead of waiting for file processing."
             (treemacs-run-in-every-buffer
              (treemacs--on-rename old-name new-name))
             (treemacs--set-refresh-flags (treemacs--nearest-parent-directory old-name) 'deleted old-name)
-            (treemacs--set-refresh-flags (treemacs--nearest-parent-directory new-name) 'created new-name))
+            (when (--none? (funcall it (treemacs--filename new-name) new-name) treemacs-ignored-file-predicates)
+              (treemacs--set-refresh-flags (treemacs--nearest-parent-directory new-name) 'created new-name)))
         (treemacs--set-refresh-flags (treemacs--parent path) event-type path)))))
 
 (define-inline treemacs--do-process-file-events ()


src/elisp/treemacs-filewatch-mode.el
```diff
@@ -181,7 +181,8 @@ file from caches if it has been deleted instead of waiting for file processing."
             (treemacs-run-in-every-buffer
              (treemacs--on-rename old-name new-name))
             (treemacs--set-refresh-flags (treemacs--nearest-parent-directory old-name) 'deleted old-name)
-            (treemacs--set-refresh-flags (treemacs--nearest-parent-directory new-name) 'created new-name))
+            (when (--none? (funcall it (treemacs--filename new-name) new-name) treemacs-ignored-file-predicates)
```
Alexander-Miller: Strange rename event indeed. I would've thought that emacs creates that backup file instead of renaming things.

At any rate this sort of thing should be caught by `treemacs--is-event-relevant?`. You'll need to check for `(eq action 'renamed)` and then extract and test the new name.

A file's parent should be fetched with `treemacs--parent`. It looks like `treemacs--nearest-parent-directory` missed a refactoring round or two, and is basically useless. I'll slot it for removal.

BlindingDark:
@Alexander-Miller If I move code here
```elisp
(define-inline treemacs--is-event-relevant? (event)
;; ...
(not (or (eq action 'stopped)
         (and (eq action 'changed)
              (not treemacs-git-mode))
         (and (eq action 'renamed)
              (let* ((new-dir (cl-fourth ,event))
                     (new-filename (treemacs--filename new-dir)))
                (--any? (funcall it new-filename new-dir) treemacs-ignored-file-predicates)))
         (--any? (funcall it filename dir) treemacs-ignored-file-predicates)))
;; ...
```
it will skip the action of deleting old file
image

Alexander-Miller:
:+1: 

I don't like running these checks in two places like this, but you're right that there's no way to avoid this.

PR #547, opened 2019-10-08
[Git] Fix directory face override project face.

BlindingDark:
Before:
image

After:
image

### DIFF
--- a/src/elisp/treemacs-async.el
+++ b/src/elisp/treemacs-async.el
@@ -310,10 +310,11 @@ EXCLUDE-PARENTS: Boolean"
              ;; then the directories
              (pcase-dolist (`(,file . ,state) output)
                (-when-let (pos (treemacs-find-visible-node file))
-                 (-let [face (treemacs--git-status-face state 'treemacs-directory-face)]
-                   (put-text-property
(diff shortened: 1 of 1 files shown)


Alexander-Miller:
This fix should not be needed. A change in a file should lead to a re-highlight of all its parents up to, but *excluding*, the root. So this is treating the symptom, but not the cause, of the problem.

I've run into this bug as well, but most of the time things work as intended and I have yet to narrow down what's causing it. Do you have a recipe to more or less reliably reproduce this?

BlindingDark:
@Alexander-Miller 
> reliably reproduce

```shell
emacs -q
```
```elisp
(treemacs-git-mode 'deferred)
```

1. Create a new git repository.
1. Add a new file to the **root** of the project.
1. **Commit!**
1. Modify this file and save it.

image


I noticed that if git state of the file has not changed, it will not reproduce.
So, for example, commit and then modify will trigger this bug.

Alexander-Miller:
Ah, looks like the setup code is wrong when the file is directly under the root. I'll look into this.

Alexander-Miller:
I'll be fixing this as part of everything that comes with #558.

Issue #540, opened 2019-09-30
error after update to Emacs 27


slk500:
GNU Emacs 27.0.50 (build 2, x86_64-pc-linux-gnu, GTK+ Version 3.22.11) of 2019-09-28
After update while using Emacs form time to time I got this error: 
[Treemacs] "  File "/home/slk500/.emacs.d/elpa/treemacs-20190922.1026/treemacs-single-file-git-status.py", line 65
    elisp_conses = "".join([f'("{path}" . "{state}")' for path,state in result_list])

How I can fix this?

Alexander-Miller:
What's the output of `git status --porcelain --ignored .` for the project where the error happens?

slk500:
Saving file /home/slk500/Dropbox/phprefactor/index.org...
Wrote /home/slk500/Dropbox/phprefactor/index.org
[Treemacs] Update of node "/home/slk500/Dropbox/phprefactor/index.org" failed with status "exited abnormally with code 1" and result
[Treemacs] "  File "/home/slk500/.emacs.d/elpa/treemacs-20190922.1026/treemacs-single-file-git-status.py", line 65
    elisp_conses = "".join([f'("{path}" . "{state}")' for path,state in result_list])

git status --porcelain --ignored .
 D index.html
 M index.org
?? clean.html~
?? clean.org~
?? index.html~
?? index.org~
?? phpunithack.sh
?? test.php

Alexander-Miller:
Locate treemacs' installation directory and run the status script like this:

`python -O treemacs-single-file-git-status.py FILE .  PARENTS` 

where FILE is a single file that's been changed and PARENTS is a list of the file's parent directories up to, and excluding, the project root.

It should look like this:

```
[I]-(a@a-laptop)-(/home/a/Documents/git/treemacs)-(master)
❯❯❯ python -O src/scripts/treemacs-single-file-git-status.py /home/a/Documents/git/treemacs/src/elisp/treemacs-core-utils.el . /home/a/Documents/git/treemacs/src/elisp /home/a/Documents/git/treemacs/src
(("/home/a/Documents/git/treemacs/src/elisp/treemacs-core-utils.el" . "M")("/home/a/Documents/git/treemacs/src/elisp" . "M")("/home/a/Documents/git/treemacs/src" . "M"))
```

Best use the index.html. Treemacs doesn't handle the deleted status, so maybe that's the cause of the trouble.

jclosure:
Same when I run the status script issue:
[...]

PR #548, opened 2019-10-09
[Git] Fix syntax error for Python3.* < Python3.6

jclosure:
Addresses issue #540 

PEP 498: Formatted string literals were not available until
Python3.6.  Adjusted `treemacs-single-file-git-status.py`
to use generally available interpolation formatting.

### DIFF
--- a/src/scripts/treemacs-single-file-git-status.py
+++ b/src/scripts/treemacs-single-file-git-status.py
@@ -62,8 +62,8 @@ def main():
             result_list.append((proc_list[i][0], propagate_state))
             i += 1
 
-    elisp_conses = "".join([f'("{path}" . "{state}")' for path,state in result_list])
-    elisp_alist = f"({elisp_conses})"
+    elisp_conses = "".join(['("{}" . "{}")'.format(path, state) for path,state in result_list])
+    elisp_alist = "({})".format(elisp_conses)
     print(elisp_alist)
 
 def add_git_processes(status_listings, path):


jclosure:
btw, looks like the travis-ci build is failing because:
```
  - Installing [16/16] evil (latest)... downloadingDependency evil failed to install: Package ‘undo-tree-0.6.3’ is unavailable
Output:
nil
The command "cask install --verbose" failed and exited with 255 during .
Your build has been stopped.
```
This is because the undo-tree available from the gnu package repo, is unsigned.  Most people have been upgrading to Emacs 26.3 to work around this: https://github.com/bbatsov/prelude/issues/1225

Alexander-Miller:
:+1: 

> This is because the undo-tree available from the gnu package repo, is unsigned.

Ah, so that's what's been going on here. I'd prefer not to change my build too much because of this, do you know if it's possible to tell cask to use this workaround?

> If you are seeing this issue, adding this temporary fix at the beginning of init.el may help:
> (setq gnutls-algorithm-priority "NORMAL:-VERS-TLS1.3")

src/scripts/treemacs-single-file-git-status.py
```diff
@@ -62,8 +62,8 @@ def main():
             result_list.append((proc_list[i][0], propagate_state))
             i += 1
 
-    elisp_conses = "".join([f'("{path}" . "{state}")' for path,state in result_list])
-    elisp_alist = f"({elisp_conses})"
+    elisp_conses = "".join(['("{}" . "{}")'.format(path, state) for path,state in result_list])
```
Alexander-Miller: The commit message should be prefixed with `[Git]`.

Alexander-Miller:
:+1: 

FYI you should bump tickets with a comment when you update, I don't get notifications about pushes.

Issue #550, opened 2019-10-11
many python processes of treemacs-single-file-git-status.py that don't complete 


jclosure:
When I modify a file in a project defined in treemacs, it does not update the color.  Instead it spawns a python process for `treemacs-single-file-git-status.py` that never completes.  I get many of these for every file changed.  

```
Treemacs Upd... 17426   run     --              --           /usr/bin/python3 -O /home/me/.emacs.d/elpa/treemacs-20190922.1026/treemacs-single-file-git-status.py /home/me/project/Dockerfile 0 /home/me/project
Treemacs Upd... 21258   run     --              --           /usr/bin/python3 -O /home/me/.emacs.d/elpa/treemacs-20190922.1026/treemacs-single-file-git-status.py /home/me/project/Dockerfile 0 /home/me/project
Treemacs Upd... 22394   run     --              --           /usr/bin/python3 -O /home/me/.emacs.d/elpa/treemacs-20190922.1026/treemacs-single-file-git-status.py /home/me/project/Dockerfile 0 /home/me/project
Treemacs Upd... 24315   run     --              --           /usr/bin/python3 -O /home/me/.emacs.d/elpa/treemacs-20190922.1026/treemacs-single-file-git-status.py /home/me/project/Dockerfile 0 /home/me/project
```

There is also a buffer created for each of these processes that just hangs around.

```
Treemacs Update Single File Process
Treemacs Update Single File Process<2> 
Treemacs Update Single File Process<3>
…
```

I notice when I run it manually, it does the same thing as above, new process and new buffer that never completes/closes.

```elisp
(with-selected-window (treemacs-get-local-window)
  (treemacs-update-single-file-git-state "/home/me/project/Dockerfile"))
```

This happens for both `(treemacs-git-mode 'deferred)` and `(treemacs-git-mode 'extended)`.  Any idea why?

I'm running Python3.7.  It happens on multiple machines.  Here's my config
```elisp
(use-package treemacs
  :ensure t
  :defer 1
  :config
  (progn
    (setq treemacs-be-evil                       nil
          treemacs-collapse-dirs                 (if (executable-find "python3") 3 0)
          treemacs-deferred-git-apply-delay      0.5
          treemacs-display-in-side-window        t
          treemacs-eldoc-display                 t
          treemacs-file-event-delay              2000
          treemacs-max-git-entries               10000
          treemacs-file-follow-delay             0.2
          treemacs-follow-after-init             nil
[...]

PR #551, opened 2019-10-12
[Git] Fix deadlock in git status python subprocesses

jclosure:
[Git] Addresses issue: https://github.com/Alexander-Miller/treemacs/issues/550

Deadlock can happen when large amounts of data are returned from
git commands inside Popen() with stdout=PIPE.

Relevant reference can be found here:
https://docs.python.org/3/library/subprocess.html#subprocess.Popen.wait

Adjusting to ensure that treemacs can handle larger projects that trigger this condition by avoiding the deadlock condition all together.


Alexander-Miller:
:+1:

PR #557, opened 2019-11-03
Use run-hook-with-args instead of run-hook-wrapped

yyoncho:


Alexander-Miller:
:+1: 

Whyever did I pick run-wrapped in the first place ...

PR #577, opened 2019-12-04
Fix the sorting order

Vifon:
The ascending/descending orders seemed to be reversed.

### DIFF
--- a/src/elisp/treemacs-customization.el
+++ b/src/elisp/treemacs-customization.el
@@ -209,7 +209,7 @@ To keep the alist clean changes should not be made directly, but with
   :type 'regexp
   :group 'treemacs)
 
-(defcustom treemacs-sorting 'alphabetic-desc
+(defcustom treemacs-sorting 'alphabetic-asc
   "Indicates how treemacs will sort its files and directories.
 Files will still always be shown after directories.
 
--- a/src/elisp/treemacs-rendering.el
+++ b/src/elisp/treemacs-rendering.el
@@ -97,53 +97,53 @@ the height of treemacs' icons must be taken into account."
   "Sort F1 and F2 alphabetically asc."
   (declare (pure t) (side-effect-free t))
   (inline-letevals (f1 f2)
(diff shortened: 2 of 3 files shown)


src/elisp/treemacs-customization.el
```diff
@@ -209,7 +209,7 @@ To keep the alist clean changes should not be made directly, but with
   :type 'regexp
   :group 'treemacs)
 
-(defcustom treemacs-sorting 'alphabetic-desc
+(defcustom treemacs-sorting 'alphabetic-asc
```
Alexander-Miller: Installation snippet in the readme should be updated as well.

src/elisp/treemacs-rendering.el
```diff
@@ -97,53 +97,53 @@ the height of treemacs' icons must be taken into account."
   "Sort F1 and F2 alphabetically asc."
   (declare (pure t) (side-effect-free t))
   (inline-letevals (f1 f2)
-    (inline-quote (string-lessp ,f2 ,f1))))
```
Alexander-Miller: Commits should be squashed, commit message should be prefixed with `[Rendering]`.

src/elisp/treemacs-rendering.el
```diff
@@ -97,53 +97,53 @@ the height of treemacs' icons must be taken into account."
...
+    (inline-quote (string-lessp ,f1 ,f2))))
```
Alexander-Miller: Is `make test` still green?

src/elisp/treemacs-customization.el
Vifon: Done.

src/elisp/treemacs-rendering.el
Vifon: Done.

src/elisp/treemacs-rendering.el
Vifon: Yes, it is.

Alexander-Miller:
:+1:

PR #580, opened 2019-12-05
[Compatibility] update winum regex as scoped buffers were added

mpanarin:
`Treemacs-Scoped-Buffer`s were added, so it is better to ignore any buffer containing name prefix


Alexander-Miller:
Thanks, but I already had my own fix in the pipeline.

PR #587, opened 2019-12-11
[Rendering] Add directory / file name transformers

### LINKED ISSUE #584: Is there a way to hide file extensions?
I have a project that only contains org files so I think it would look way cleaner without file extensions, like so:
```
- Personal
 Main
 Auxiliary
```
rather than the way it is now:
```
- Personal
 Main.org
 Auxiliary.org
```

EDIT: If this is not currently possible I'd appreciate if you could point out the function that handles node rendering, thanks.

TonCherAmi:
Hopefully I haven't missed anything.

resolves #584


Alexander-Miller:
:+1:

PR #591, opened 2019-12-20
[Misc] Verify that shelf is not nil before using it

yyoncho:
- it breaks extensions (e. g. lsp-treemacs)


Alexander-Miller:
:+1:

Issue #597, opened 2019-12-31
Summary of treemacs-persp package looks wrong


kamoii:
![screenshot-2019-12-31T17:30:54](https://user-images.githubusercontent.com/52167401/71615228-86647800-2bf3-11ea-8b40-b51fd4b74fde.png)

PR #599, opened 2020-01-03
[Persp] Update persp-mode integration summary

nixmaniack:
Fix https://github.com/Alexander-Miller/treemacs/issues/597


Alexander-Miller:
:+1:

PR #611, opened 2020-01-25
[Logging] Show error when Changelog.org not found

Compro-Prasad:


Alexander-Miller:
Oof. Thanks for the fix, but that is not supposed to be necessary. Looks like I have simply forgotten to include it in the melpa recipe, and no one ever noticed, or at least they didn't complain. 
I'll close here because the change log should always be present and it changelog should not be a simple oopsie that we log and forget, it's an error and should be visible as such. Patch for the recipe will follow shortly.

Alexander-Miller:
https://github.com/melpa/melpa/pull/6681

Took a bit longer than intended, but that PR should fix the missing changelog.

PR #612, opened 2020-01-25
Refine document.

zw963:


README.org
```diff
@@ -158,7 +158,7 @@ The used org-format is quite simple: level 1 headlines are names of workspaces,
 in a workspace, and every project's path is given as a description list, starting with a ~-~ (and an optional leading
 space). Empty lines and lines starting with ~#~ are ignored, and everything else leads to an error.
 
-You needn't worry about making mistakes either. If there's something wrong when you call ~treemacs-finish-edit~ then
+You needn't worry about making mistakes either. If there's something wrong when you call ~treemacs-finish-edit~ (C-c C-c) then
```
Alexander-Miller: The line's too long. You need to call `fill-paragraph`.

Alexander-Miller: The commit message should be prefixed with "[Docs]". It also also too generic, let's go with something like "Mention finish-edit keybind in workspace edit section".

zw963: Thank you for code review, updated.

Alexander-Miller:
:+1:

PR #628, opened 2020-03-01
Add support for custom mode line format

caseneuve:
Allows user to specify their `treemacs` mode-line format through setting variable `treemacs--user-mode-line-format` to desired value without overriding `treemacs--setup-mode-line` function.

### DIFF
--- a/src/elisp/treemacs-mode.el
+++ b/src/elisp/treemacs-mode.el
@@ -353,7 +353,9 @@ to it will instead show a blank."
 (defun treemacs--setup-mode-line ()
   "Create either a simple modeline, or integrate into spaceline."
   (setq mode-line-format
-        (cond ((fboundp 'spaceline-install)
+        (cond (treemacs-user-mode-line-format
+               treemacs-user-mode-line-format)
+              ((fboundp 'spaceline-install)
                (spaceline-install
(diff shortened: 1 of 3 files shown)


src/elisp/treemacs-mode.el
```diff
@@ -350,10 +350,15 @@ to it will instead show a blank."
       map)
     "Keymap for `treemacs-mode'."))
 
+(defvar treemacs--user-mode-line-format nil
```
Alexander-Miller: Since this is a user-facing customization variable it should be placed in treemacs-customization.el.

Being public it also shouldn't have a double-dash.

Finally it must be mentioned in the readme both in the [install snippet](https://github.com/Alexander-Miller/treemacs#installation) and the [config option table](https://github.com/Alexander-Miller/treemacs#variables).

caseneuve: Thanks for the review (and sorry I overlooked the README.org update). I hope it's better now.

Alexander-Miller: LGTM. Squash it down to a single commit and we're good to merge.

caseneuve: Done

Alexander-Miller:
:+1:

PR #640, opened 2020-04-02
Make modes with treemacs icons configurable

yyoncho:


src/elisp/treemacs-themes.el
```diff
@@ -38,6 +38,9 @@
 
 (defvar treemacs--themes nil "List of all known themes.")
 
+(defvar treemacs-modes '(treemacs-mode dired-mode)
```
Alexander-Miller, line 41: Let's call this `treemacs-icon-modes` so it's clear which modes this is about.

The commit message should also be start with `[Icons]`.

yyoncho:
Turns out I don't need that.

Alexander-Miller:
:man_shrugging:

PR #648, opened 2020-04-14
Fix minor typo in user mode line format variable

mcraveiro:
Second take of the PR with commit under ```[Docs]```.


Alexander-Miller:
:+1:

PR #670, opened 2020-05-08
[Dired] Add missing "require" for treemacs-logging.

leungbk:


Alexander-Miller:
Thanks, but this change is not necessary. icons-dired is its own independent package that depends on treemacs - *all* of treemacs. The fact that there's a logging module is an implementation detail it does not need to care about.

PR #673, opened 2020-05-10
[Icons] Enable icon resizing on Emacs 27 or newer

casouri:
Even if ImageMagick is not available. Because Emacs 27 now comes with native image scaling.


casouri:
And I just see the other PR. Nevermind.

PR #675, opened 2020-05-16
[Header Line] Hide global header line if there is one

jerrypnz:
Make sure buffer-local `header-line-format` is set to `nil` when
`treemacs-user-header-line-format` is set to `nil`. This change is
useful for those who use a header line instead of a mode line. When no
custom header is set for treemacs, it now does not show the global
one.

Before the change:

Screen Shot 2020-05-16 at 10 30 46 PM

After the change:

Screen Shot 2020-05-16 at 10 31 41 PM


Alexander-Miller:
:+1:

PR #702, opened 2020-07-29
Add all-the-icons theme

ericdallo:
Add a new theme using `all-the-icons`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,6 +43,7 @@ ELS += $(SRC_DIR)/treemacs-tags.el
 ELS += $(SRC_DIR)/treemacs-themes.el
 ELS += $(SRC_DIR)/treemacs-visuals.el
 ELS += $(SRC_DIR)/treemacs-workspaces.el
+ELS += $(EXTRA_DIR)/treemacs-all-the-icons.el
 ELS += $(EXTRA_DIR)/treemacs-evil.el
 ELS += $(EXTRA_DIR)/treemacs-icons-dired.el
 ELS += $(EXTRA_DIR)/treemacs-persp.el
--- a/src/elisp/treemacs-themes.el
+++ b/src/elisp/treemacs-themes.el
@@ -61,6 +61,7 @@
 (cl-defmacro treemacs-create-theme (name &key icon-directory extends config)
   "Create a new (bare) theme with the given NAME.
 - ICON-DIRECTORY is the (mandatory) theme's location.
+- EXTENDS is the theme to be extended.
 - BASED-ON is the name of a theme whose icons this one should start with.
 - CONFIG is a code block to fill the created theme with icons via
   `treemacs-create-icon'."
--- a/src/extra/treemacs-all-the-icons.el
+++ b/src/extra/treemacs-all-the-icons.el
@@ -1,0 +1,251 @@
+;;; treemacs-all-the-icons.el --- all-the-icons integration for treemacs -*- lexical-binding: t -*-
+
+;; Copyright (C) 2020 Alexander Miller
+
+;; Author: Eric Dallo <ercdll1337@gmail.com>
+;; Package-Requires: ((emacs "25") (all-the-icons "4.0.1") (treemacs "0.0"))
+;; Version: 0
+;; Homepage: https://github.com/Alexander-Miller/treemacs
+
+;; This program is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+;;; Commentary:
+;;; all-the-icons integration.
+;;
+;;; Code:
+
+(require 'all-the-icons)
+(require 'treemacs)
+
+(defface treemacs-all-the-icons-root-face
+  '((t (:inherit font-lock-string-face)))
+  "Face used for the root icon in all-the-icons theme."
+  :group 'treemacs-faces)
+
+(defface treemacs-all-the-icons-file-face
+  '((t (:inherit font-lock-doc-face)))
+  "Face used for the directory and file icons in all-the-icons theme."
+  :group 'treemacs-faces)
+
+(treemacs-create-theme "all-the-icons"
+  :config
+  (progn
+    ;; directory and other icons
+    (treemacs-create-icon :icon (format " %s\t" (all-the-icons-octicon "repo" :height 1.2 :v-adjust -0.1 :face 'treemacs-all-the-icons-root-face))
+                          :extensions (root)
+                          :fallback "")
+    (treemacs-create-icon :icon (format "%s\t%s\t"
+                                        (all-the-icons-octicon "chevron-down" :height 0.75 :v-adjust 0.1 :face 'treemacs-all-the-icons-file-face)
+                                        (all-the-icons-octicon "file-directory" :v-adjust 0 :face 'treemacs-all-the-icons-file-face))
+                          :extensions (dir-open)
+                          :fallback (propertize "+ " 'face 'treemacs-term-node-face))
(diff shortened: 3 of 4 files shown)


src/elisp/treemacs-icons.el
```diff
@@ -294,7 +294,6 @@ Necessary since root icons are not rectangular."
     (treemacs-create-icon :file "asciidoc.png"      :extensions ("adoc" "asciidoc"))
     (treemacs-create-icon :file "rust.png"          :extensions ("rs"))
     (treemacs-create-icon :file "image.png"         :extensions ("jpg" "jpeg" "bmp" "svg" "png" "xpm" "gif"))
-    (treemacs-create-icon :file "emacs.png"         :extensions ("el" "elc"))
```
Alexander-Miller, line 297: That shouldn't happen.

src/elisp/treemacs-icons.el
```diff
@@ -426,6 +425,221 @@ Necessary since root icons are not rectangular."
     (treemacs-create-icon :file "vsc/r.png"         :extensions ("r"))
     (treemacs-create-icon :file "vsc/reason.png"    :extensions ("re" "rei"))))
 
+(defvar all-the-icons-icon-alist nil)
+(declare-function all-the-icons-octicon "ext:all-the-icons" t t)
+(declare-function all-the-icons-faicon "ext:all-the-icons" t t)
+(declare-function all-the-icons-material "ext:all-the-icons" t t)
+(declare-function all-the-icons-alltheicon "ext:all-the-icons" t t)
+
+(when (require 'all-the-icons nil t)
```
Alexander-Miller, line 434: Since we require a hard dependency to another package the whole theme should become its own package. All the relevant code should be moved into its own file in src/extra. I'll take care of publishing on melpa once this is merged.

src/elisp/treemacs-themes.el
```diff
@@ -61,6 +61,7 @@
 (cl-defmacro treemacs-create-theme (name &key icon-directory extends config)
   "Create a new (bare) theme with the given NAME.
 - ICON-DIRECTORY is the (mandatory) theme's location.
+- EXTENDS is the theme to be extended.
```
Alexander-Miller, line 64: Commits need to be squashed, the commit message should be prefixed with `[Themes]`.

Only this change should remain its own commit since it's unrelated to the larger theme addition.

ericdallo: Ok!

src/elisp/treemacs-icons.el
ericdallo: I see, I'll move it!

src/elisp/treemacs-icons.el
ericdallo: Ops, I'll undo that

ericdallo:
Done @Alexander-Miller

src/elisp/treemacs-faces.el
```diff
@@ -35,6 +35,16 @@ Using this face is incompatible with `treemacs-git-mode' (exept for the simple
 variant), so it will only be used if git-mode is disabled or set to simple."
   :group 'treemacs-faces)
 
+(defface treemacs-all-the-icons-root-face
```
Alexander-Miller, line 38: Faces should be in theme file as well.

src/elisp/treemacs-faces.el
```diff
@@ -35,6 +35,16 @@ Using this face is incompatible with `treemacs-git-mode' (exept for the simple
...
+  '((t (:inherit font-lock-string-face)))
+  "Face used for the root icon in all-the-icons theme."
+  :group 'treemacs-faces)
+
+(defface treemacs-all-the-icons-file-face
+  '((t (:inherit font-lock-doc-face :slant normal)))
```
Alexander-Miller, line 44: The slant looks like an adjustment for your own theme, so I don't think it should be a general setting.

Besides, shouldn't this inherit `treemacs-file-face` instead?

src/extra/treemacs-all-the-icons.el
```diff
@@ -0,0 +1,240 @@
+;;; treemacs-all-the-icons.el --- all-the-icons integration for treemacs -*- lexical-binding: t -*-
+
+;; Copyright (C) 2020 Alexander Miller
+
+;; Author: Alexander Miller <alexanderm@web.de>
```
Alexander-Miller, line 5: Don't know what the proper etiquette about these things is like, but you wrote (the first version of) this theme, so IMO  you should put yourself down as the author as well. At least that's how it went in treemacs-perspective.el.

src/extra/treemacs-all-the-icons.el
```diff
@@ -0,0 +1,240 @@
...
+                          :extensions ("class" "exe" "dll" "obj" "so" "o" "out" "elc" "pyc"))
+    (treemacs-create-icon :icon (format "  %s\t" (all-the-icons-octicon "file-pdf" :v-adjust 0 :face 'all-the-icons-orange))
+                          :extensions ("pdf"))
+    (treemacs-create-icon :icon (format "  %s\t" (all-the-icons-octicon "file-zip" :v-adjust 0 :face 'treemacs-all-the-icons-file-face))
+                          :extensions ("zip" "7z" "tar" "gz" "rar" "tgz"))
+
+    (treemacs-create-icon :icon (format "  %s\t" (all-the-icons-octicon "file-text" :v-adjust 0 :face 'treemacs-all-the-icons-file-face))
+                          :extensions (fallback))))
+
+(provide 'treemacs-all-the-icons)
```
Alexander-Miller, line 239: I always leave an empty line between provide and end comment.

Alexander-Miller:
Tabs become larger when you expand a directory. That's just me, right?
tab

src/elisp/treemacs-faces.el
ericdallo: Removed the `slant`.
I could not make it work inheriting from `treemacs-file-face`, maybe all-the-icons face needs foreground and background declared on the face?

src/extra/treemacs-all-the-icons.el
ericdallo: I had the same doubt :sweat_smile:, I think it makes sense.

ericdallo:
@Alexander-Miller that does not happen to me:
image

Alexander-Miller:
And now you see why I've never gotten into all-the-icons - it's never looked right for me.

At any rate I consider that a local problem, it's not an impediment for merging this PR.

src/elisp/treemacs-faces.el
Alexander-Miller: Strange, it works on my end.

Using doc-face here is for the best anyway, treemacs-file-face just inherits default, and icons should have some color.

ericdallo:
I believe that problem happens because of the tab space, your tab space may be different from mine (2 space, 4 spaces etc).

src/extra/treemacs-all-the-icons.el
```diff
@@ -0,0 +1,251 @@
+;;; treemacs-all-the-icons.el --- all-the-icons integration for treemacs -*- lexical-binding: t -*-
+
+;; Copyright (C) 2020 Eric Dallo
```
Alexander-Miller, line 3: All right, one final thing. For technical correctness I think this part needs to go back to my name - it'll be merged in my package and I'll be responsible for maintaining it in the future.

Makefile
```diff
@@ -15,6 +15,7 @@ LINT_FLAG        = --eval "(setq byte-compile-dest-file-function (lambda (f) (co
 TEST_COMMAND     = buttercup -L $(SRC_DIR) -L $(EXTRA_DIR) $(NO_LOAD_WARNINGS)
 
 ELS  = $(SRC_DIR)/treemacs.el
+ELS  = $(SRC_DIR)/treemacs-all-the-icons.el
```
Alexander-Miller, line 18: Never mind, copyright's not quite the last thing. This line needs to use `+=` and `SRC_DIR` and should be moved down to the other addons in line 50.

Alexander-Miller:
Tab width is not the cause, I've tried everything from 1 to 4 and the behavior is always the same.

src/extra/treemacs-all-the-icons.el
Alexander-Miller: Hold on, you reverted too far. I only meant the copyright, you should still be the author.

ericdallo: Oh, sorry, I see

ericdallo: Done!

Alexander-Miller:
Next stop: melpa.

ericdallo:
Thank you!

PR #715, opened 2020-08-27
Fix fringe customization

sebastiencs:
- Fix display property of the fringe overlay 
  This worked because the defcustom and the bitmap were on the same
  symbol.
  The fringe property is expecting the symbol of the bitmap, so this
  ignored the value of the defcustom.

- Make the default fringe bitmap bigger and fix type of the defcustom
  The default bitmap didn't fill the whole line in my config


Alexander-Miller (changes requested):
LGTM. All that's missing is to prefix the commit messages with `[Fringe Indicator]`.

sebastiencs:
@Alexander-Miller Done

Alexander-Miller (approved):

Alexander-Miller:
:+1:

PR #719, opened 2020-09-16
Fix for  jorgenschaefer/emacs-buttercup#198

snogge:
-L . is required to find the test files.

Fixes jorgenschaefer/emacs-buttercup#198


Alexander-Miller:
Thanks :+1:

Issue #725, opened 2020-09-29
Typo


zw963:
![image](https://user-images.githubusercontent.com/549126/94566161-9e5d4b00-029c-11eb-991d-92f0f6910b31.png)

SanjayDevTech:
Shall I fix this typo @zw963 ?
Could you assign this issue to me

zw963:
@SanjayDevTech , i think i can't assign this issue to someone. no permission.

Alexander-Miller:
I don't do any ticket assigning. Dropping an FYI if you want to contribute is enough.

PR #728, opened 2020-09-30
Fix typo

zw963:
#725

### DIFF
--- a/src/elisp/treemacs-header-line.el
+++ b/src/elisp/treemacs-header-line.el
@@ -67,12 +67,12 @@
       (let* ((menu
               (easy-menu-create-menu
                nil
-               `(["Edit Workspaces"       treemacs-edit-workspaces]
-                 ["Create Workspace"      treemacs-create-workspace]
-                 ["Remove Worspace"       treemacs-remove-workspace]
-                 ["Rename Workspace"      treemacs-rename-workspace]
-                 ["Switch Worspaces"      treemacs-switch-workspace]
-                 ["Set Fallback Worspace" treemacs-set-fallback-workspace])))
+               `(["Edit Workspaces"        treemacs-edit-workspaces]
+                 ["Create Workspace"       treemacs-create-workspace]
+                 ["Remove Workspace"       treemacs-remove-workspace]
+                 ["Rename Workspace"       treemacs-rename-workspace]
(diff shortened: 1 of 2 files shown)


src/elisp/treemacs-header-line.el
```diff
@@ -69,10 +69,10 @@
                nil
                `(["Edit Workspaces"       treemacs-edit-workspaces]
                  ["Create Workspace"      treemacs-create-workspace]
-                 ["Remove Worspace"       treemacs-remove-workspace]
```
Alexander-Miller, line 72: The values aligning as before is more important than a minimal diff.

src/elisp/treemacs-header-line.el
```diff
@@ -69,10 +69,10 @@
...
+                 ["Remove Workspace"       treemacs-remove-workspace]
```
Alexander-Miller, line 72: The commit message should be prefixed with `[Misc]`.

Alexander-Miller (changes requested):

src/elisp/treemacs-header-line.el
zw963: Fixed.

Alexander-Miller:
:+1:

src/elisp/treemacs-mouse-interface.el
```diff
@@ -242,12 +242,12 @@ and ignore any prefix argument."
...
                 ("Workspaces"
-                 ["Edit Workspaces"       treemacs-edit-workspaces]
-                 ["Create Workspace"      treemacs-create-workspace]
-                 ["Remove Worspace"       treemacs-remove-workspace]
-                 ["Rename Workspace"      treemacs-rename-workspace]
-                 ["Switch Worspaces"      treemacs-switch-workspace]
-                 ["Set Fallback Worspace" treemacs-set-fallback-workspace])
+                 ["Edit Workspaces"        treemacs-edit-workspaces]
+                 ["Create Workspace"       treemacs-create-workspace]
+                 ["Remove Worskpace"       treemacs-remove-workspace]
```
chessman, line 247: that's another typo

Alexander-Miller: Doh. 

Fixing it now.

PR #738, opened 2020-11-05
simplify help code, remove hydra dependency

minad:
Instead of generating a hydra, simply show the help string via the `lv` package. There is no need to generate a hydra on the fly. The only difference is that we have to introduce a `treemacs-help-key-face`. But this is ok I guess, since there are already other `treemacs-help-faces`.

**Edit**: There is actually one significant difference. The hydra closes the help if you press a key which is not present in the transient keymap of the hydra. This won't work here, you have to close the help explicitly. Maybe you like the hydra behavior better?

**Edit 2**: Turns out the hydra behavior can easily be emulated by `set-transient-map`. So now it should really be equivalent in behavior.

**Edit 3** Turns out, setting the transient map is not good enough if we are calling for example the project rename command. I guess this is not worth further efforts...


Alexander-Miller:
Well that was a roller coaster lol

Fortunately hydra is a lazy dependency, it's not loaded until actually used. It's also a popular package, so having treemacs depend on it does not cost us much.

minad:
Yes, I used hydra for quite some time and like it very much. Right now I am experimenting with my own small hydra-like implementation, which is a bit adapted to my needs (or in other words, I am playing around). So I was wondering if it is necessary for treemacs to require a hard dependency on hydra. As I see now if you want the help hydra to behave as is - there is not really a way around it, if you want a well behaved transient keymap.

The only option to avoid hydra here is if we accept that foreign keys don't close the help view and it just stays open until the user presses "?". Then this PR would indeed work out and make things simpler, since no "defhydra" is needed.

PR #758, opened 2021-01-03
fix: treemacs-perspective misspelling as treemacs-persective

Randy1Burrell:
treemacs-perspective was spelt wrong so when I copied the spelling from the
README file and pasted it into my configuration to install, I got an error
stating that no package exists in Melpa by that name. I corrected this just
in-case someone else does the same thing in the future, then there won't be any
errors.


Alexander-Miller:
:+1:

PR #759, opened 2021-01-05
Fix unknown change event: `magit-refresh`

nbfalcon:
`treemacs-magit--schedule-refresh` calls `treemacs--set-refresh-flags` with
`magit-refresh`, which, being a change event not explicitly handled, causes the
above warning. The only side effect of the call to `treemacs--set-refresh-flag`
is to cause the `treemacs` project to be refreshed, though, so call
`treemacs-project->refresh` directly, fixing the message.

----

#

### DIFF
--- a/src/elisp/treemacs-filewatch-mode.el
+++ b/src/elisp/treemacs-filewatch-mode.el
@@ -62,6 +62,14 @@ buffers watching that path, its cdr is the watch descriptor.")
   "Timer that will run a refresh after `treemacs-file-event-delay' ms.
 Stored here to allow it to be cancelled by a manual refresh.")
 
+(define-inline treemacs--start-filewatch-timer ()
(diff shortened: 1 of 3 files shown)


Alexander-Miller:
Hmm, looks like you uncovered a somewhat deeper sitting bug here.

I didn't want the main package to have direct knowledge of the `magit-refresh` event, so I pushed it 8 times, because
that is the magic number where treemacs stops reacting to single changes and just force re-opens the entire directory.

However I forgot about making the event-pushing code smart enough not to accept duplicates, so only a single
`magit-refresh` is ever set.

Your solution comes close, but there's still a bit of detail work in the background that should be adjusted.

Alexander-Miller:
Ok, things should work now in cooperation with the usual filewatch timer mechanism.

Let me know if I missed anything.

PR #766, opened 2021-01-15
Fix Perspective compatibility

gcv:
Here from https://github.com/nex3/perspective-el/issues/144

The treemacs-perspective wrapper calls `persp-killed-hook` incorrectly. This hook in Perspective does not have a "scope" or "name" parameter. The perspective being killed is the current perspective, and the usual introspection helpers for determining the current perspective are available.

Thanks to @Randy1Burrell for reporting the bug and testing the patch.


Alexander-Miller:
:+1:

PR #771, opened 2021-01-31
[Misc] Align some slashes for Clippy

MageJohn:
The slashes for the tail of Clippy's speech bubble were one on top of
the other, like this:
\\
\\

This is clearly the most important PR you've ever received :P


Alexander-Miller:
I don't think the original looks all that bad, but I like yours better :+1:

PR #774, opened 2021-02-03
[Doc] Update examples to adhere to use-package semantics

Randy1Burrell:
In the `use-package` documentation it suggested that when using the `:after`
keyword with multiple packages, they should be in a list. This have become
apparent to me after working with `treemacs` for sometime and getting certain
errors. I updated the example in the doc, just in-case other users copy the
examples and try to use it directly without modification, then it will just work
without problems.

If applied, this commit will:

    - Update the example that uses use-package :after key word with multiple
      packages to be a list.

### DIFF
--- a/README.org
+++ b/README.org
@@ -457,24 +457,24 @@ setting. Setting them all yourself is not necessary, they are only listed here t
           ("C-x t M-t" . treemacs-find-tag)))
 
   (use-package treemacs-evil
-    :after treemacs evil
+    :after (treemacs evil)
     :ensure t)
 
   (use-package treemacs-projectile
-    :after treemacs projectile
+    :after (treemacs projectile)
     :ensure t)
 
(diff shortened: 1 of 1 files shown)


Alexander-Miller:
LGTM. Update the prefix to `[Docs]` and we can merge.

Randy1Burrell:
What prefix? Are you talking about the prefix in the commit message?

Alexander-Miller:
Yes, I meant the commit message. Makes it easier for me to look things up and the [Format] gets special syntax highlighting in magit.

Randy1Burrell:
That should do it.

Alexander-Miller:
Nitpick: you're missing an "s" for the "Docs".

Randy1Burrell:
Sorry about that. I must be too tired. It's updated now.

Alexander-Miller:
:+1:

PR #779, opened 2021-02-25
Fix reading treemacs-persist-file containing multibyte symbols

axard:
I found one bug. If I use multibyte symbols in workspace name or path it renders like that
treemacs_bug
The reason of such behavior in using `insert-file-contents-literally` which doesn't decodes text from file.


Alexander-Miller:
LGTM. Prefix the commit message with `[Persistence]` and we can merge.

Alexander-Miller:
:+1: 

(Note: you should say when you push a new version, there's no notification for new commits.)

PR #784, opened 2021-03-04
Fix treemacs--file-name-handler-alist for TRAMP

prashantvithani:
When 6c431174dd9 introduced treemacs--file-name-handler-alist, it
mistakenly used the wrong format for alists. This broke TRAMP
integration, because the `file-name-handler-alist` was no longer a valid
alist. Fortunately, this is a trivial fix.


Alexander-Miller:
:+1:

PR #785, opened 2021-03-05
[Icon] .gitattributes file should share the same icon with .gitignore

rayw000:
`.gitattributes` file should also share the same icon as `.gitignore` in `all-the-icons` integration.


Alexander-Miller:
:+1:

Issue #677, opened 2020-05-19
Towards a polymorphic extensions implementation


Alexander-Miller:
The extension api is due for an overhaul. It contains many obsolete atavisms that complicate the implementation and by now can be safely deleted. The biggest problem however is the use of macros to generate many specialized functions which greatly complicates development, debugging and introspection.

Fortunately the problem space is ideal for an OO-based approach - all the different node types are all just different implementations of the same behavior, declaring their icons, faces query-functions etc. Tracking state is not necessary either, so using static functions will be well enough.

I see 2 possible downsides that must be considered:

Performance: eieio method calls are about 4 times slower than normal function calls. Should this have an actual impact on performance a workaround can be applied, albeit at the cost of slightly more complex architecture: the node objects methods, instead of directly producing values like their faces, will need to instead output a function. That will eliminate performace concerns, since using `funcall` is just as fast as a normal function call.

Backward compatibility: Most of the rebuild should focus on interior implementation details, but if breakage does turn out to be unavoidable it's probably best to publish the new implementation in a new module with a different name (and delete the old one some time down the road), so packages using the extension api can upgrade whenever they are ready.

Alexander-Miller:
ping @yyoncho As a heavy user of extensions I am sure you have plenty of ideas how things can be improved.

yyoncho:
I personally would vote for a data-driven solution in which the model is readable. We have created a wrapper providing this functionality. The model is just a plist with :icon/:label/:children/:children-async, and related keys. This probably will lead to slower access to the fields, but this is not an issue for us.

Alexander-Miller:
Would that plist have to be descriptive or prescriptive? If it's the former I can produce one as a side effect, or even build something interactive that pretty prints all available info.

yyoncho:
I am not sure I understand the difference between the two in this context, here it a runnable(after you load lsp-treemacs) example: 

``` emacs-lisp
(display-buffer
 (lsp-treemacs-render
  '((:key "foo"
          :label "Root"
          :icon dir-open
          :children-async (lambda (_ callback)
                            (run-with-idle-timer
                             2
                             nil
                             (lambda (callback)
                               (funcall 
                                callback 
                                '((:key "foo" 
                                        :label "Async" :icon dir-open
                                        :children ((:key "foo"
                                                         :label "Sync"
                                                         :icon dir-open
                                                         :ret-action ignore
                                                         :actions (["Right Click selected" ignore])))))))
                             callback))
[...]

PR #790, opened 2021-03-12
Fix sample extensions code

gveres:
Looks like when calling the query-function of an expandable node, the "btn" binding no longer works, but "node" does


Alexander-Miller:
:+1: 

The extension documentation is a bit neglected since the whole api is undergoing an overhaul: https://github.com/Alexander-Miller/treemacs/issues/677
If you seriously want to build something based on treemacs' extensions you should base it off the treelib branch.

PR #797, opened 2021-04-08
Populate bookmark-alist from default file

qoops-1:
On startup `bookmark-alist` is always empty, unless it's populated somewhere else, hence, calling `treemacs-bookmark` results in "Didn't find any bookmarks" error, even if bookmarks do exist. Made this quick fix that populates `bookmark-alist` inside `treemacs-bookmark` call, if needed


qoops-1:
@Alexander-Miller What do you think about it?

Alexander-Miller:
Sorry for the delay, I've been busy renovating for a few days,  just catching up now.

The change is looking good, let's get this merged :+1:

Issue #534, opened 2019-09-17
Request: Improved usability with a different set of keybindings


sdwolfz:
Here is what my general expectations would be:

- [x] `h` and `M-h` - close directory if current node is open, navigate up if current node is closed, or a file.
- [x] `l` and `M-l` - open directory if current node is closed, navigate inside to the first element if node is open, RET-action if it's not a directory
- [x] `M-S-h` (`M-H`) and `M-S-l` (`M-L`) would take the role of the current `h` and `l`. `C-h` and `C-l` are very usefull with the default functionality so that's why I choose `Meta` for the modifier.
- [ ] as @Compro-Prasad suggested, having `left` and `right` do the same thing as `h` and `l` would also be good.

The general idea is to re-allocate `h` and `l` to open/close directories, similar to whats happening in neotree. While the root manipulation keys are useful, they become useless when having multiple projects in your tree, and navigating with one hand using `h,j,k,l` while having a glass of water in the other is really something I would like more people to have the change to experience 😄 .

- [x] `Esc` when the Hydra is open should close it. This is just something that I instinctlvely press, probably due to my memory of using spacemacs transient states. I don't know a way to achieve that from my current config, without modifying the code. (https://github.com/Alexander-Miller/treemacs/pull/535)
- [ ] `'` - peek, open a file in the previous buffer, but keep the cursor in the treemacs buffer. Currently the file is only temporarily opened. Not sure how I feel about this yet, but an option to change the peek behaviour would be nice.
- [ ] `P` should probably be re-purposed for the current projects functionality, instead of using `C-c C-p`.
- [ ] `W` could also be used for workspace manipulation, instead of `C-c C-w`.
- [x] `w` for setting the width is a powerful feature, but quite awquard to use. What I suggest is to have `>` for increasing the window size by `increment` (default 1) and `<` to decrease the window size by `increment`. Instead of using the `w` for this, which could easily be mis-pressed.
- [ ] `|` and `-` to open in a split window, as suggested by @czert. Also it would be nice to have an option to invert the horizontal/vertical splits so they behave like `SPC w s` and `SPC w v` in spacemacs.

I have the following settings in my dotfiles as a workaround for the behaviour I want:
```elisp
  (with-eval-after-load "treemacs-evil"
    (evil-define-key 'treemacs treemacs-mode-map (kbd "'") #'treemacs-peek)
[...]

PR #798, opened 2021-04-08
Implement collapse action

sdwolfz:
As mentioned here: https://github.com/Alexander-Miller/treemacs/issues/534#issuecomment-810565109

This provides a `COLLAPSE` action that defined the behavior of a key that navigates up the directory structure. Together with the keybind changes allow for one handed navigation using `hjkl` (similar to nerdtree/neotree).


sdwolfz:
@Alexander-Miller should be ready for review now.

README.org
```diff
@@ -13,37 +13,37 @@
 
 * Content                                                                            :TOC:noexport:
 - [[#quick-feature-overview][Quick Feature Overview]]
-   - [[#fancy-gifs][Fancy Gifs!]]
```
Alexander-Miller, line 16: Any idea why the indent is changed? I thought setting the `org-list-indent-offset` in the header line should take care of that.

src/extra/treemacs-evil.el
```diff
@@ -97,8 +97,8 @@ Then call ORIG-FUN with its ARGS and reopen treemacs if it was open before."
 (evil-define-key 'treemacs treemacs-mode-map (kbd "gr")     #'treemacs-refresh)
 (evil-define-key 'treemacs treemacs-mode-map [down-mouse-1] #'treemacs-leftclick-action)
 (evil-define-key 'treemacs treemacs-mode-map [drag-mouse-1] #'treemacs-dragleftclick-action)
-(evil-define-key 'treemacs treemacs-mode-map (kbd "h")      #'treemacs-root-up)
-(evil-define-key 'treemacs treemacs-mode-map (kbd "l")      #'treemacs-root-down)
+(evil-define-key 'treemacs treemacs-mode-map (kbd "h")      #'treemacs-COLLAPSE-action)
+(evil-define-key 'treemacs treemacs-mode-map (kbd "l")      #'treemacs-RET-action)
```
Alexander-Miller, line 102: LGTM. Add a short entry in the Changelog.org and we're good to go.

Alexander-Miller (changes requested):

README.org
sdwolfz: No idea, I tried with both loading, and not loading settings, `toc-org-max-depth: 2` gets applied, but the offset does not... I'll edit it in fundamental mode and push.

src/extra/treemacs-evil.el
sdwolfz: Sure, btw, why does `treemacs-evil` need this particular overwrite here, but only for the `h` and `l` keys? I couldn't figure it out...

Alexander-Miller:
:+1:

src/extra/treemacs-evil.el
Alexander-Miller: treemacs-evil-state derives from motion-state, so anything already bound there needs to be overridden here.

README.org
Alexander-Miller: org-toc is always using the global value. I've opened a PR for it: https://github.com/snosov1/toc-org/pull/75

PR #799, opened 2021-04-10
I think there should be a call to `treemacs' that allows you to set focus to existing tree. There are cases when I wish to use treemacs to open a file then jump straight back with the same treemacs hotkey.

rileyrg:


rileyrg:
Error in my local merge. Sorry for the noise. I'm a bit of a beginner. Will try to rectify.

Alexander-Miller:
You're probably seeing conflicts because I force-pushed some commits that were already on master to fix a commit message.

You'll have to start a new branch on the latest origin/master and add your commit on top of that.

PR #800, opened 2021-04-11
missing treemacs call to focus existing tree if exists

rileyrg:
submitted using magit/forge - didn't prompt for a description - sorry, but pretty straightforward.

### DIFF
--- a/src/elisp/treemacs.el
+++ b/src/elisp/treemacs.el
@@ -68,16 +68,18 @@
   treemacs-version)
 
 ;;;###autoload
-(defun treemacs ()
-  "Initialise or toggle treemacs.
-* If the treemacs window is visible hide it.
+(defun treemacs (&optional focus)
+  "Initialise, focus or toggle treemacs.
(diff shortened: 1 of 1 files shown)


rileyrg:
added optional focus flag to treemacs call to focus existing tree if already open ,else create or make visible as applicable.

```
     (use-package
       treemacs
       :bind
       ("M-0"   . (lambda()(interactive)(treemacs t))))
```

Alexander-Miller:
Sorry for the late response, I've been busy renovating for a few days, just catching up now.

I don't think this is really needed when `treemacs-select-window` is also around. It has no toggle behavior and will always select the treemacs window, but that's not much of a difference, you can just press `M-0`+ `q` to quit treemacs.

If the `treemacs` command itself should have the option to select the window it should be handled with a customization variable and maybe a prefix arg that would reverse that variables behavior.

rileyrg:
There you go! I missed treemacs-select-window ...  silly me had assumed treemacs was the entry point. So fair enough, I'll close ;)

PR #805, opened 2021-04-21
Do not #'quote commands in keymap definitions

tarsius:
Function-quoting a symbol causes the byte-compiler to check whether
that symbol is defined as a function.  That is the only effect.  If
we stop doing that, then we no longer have to use `with-no-warnings`
to suppress the resulting warnings.


Alexander-Miller:
:+1:

PR #806, opened 2021-04-21
Make library commentaries comply more with the conventions.

tarsius:
- Only section headings should begin with three or more semicolons.
  Regular comments as well as all non-heading text in the library
  headers should begin with only two semicolons.

- Once the extra semicolon is removed, it looks quite ugly that there
  is no empty line between the Commentary heading and its body, so add
  that line.

- Use `fill-paragraph` on text in the library commentaries.

### DIFF
--- a/src/elisp/treemacs-async.el
+++ b/src/elisp/treemacs-async.el
@@ -16,7 +16,8 @@
 ;; along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 ;;; Commentary:
-;;; Code for dealing with asynchronous processes.
+
+;; Code for dealing with asynchronous processes.
 
 ;;; Code:
 
--- a/src/elisp/treemacs-bookmarks.el
+++ b/src/elisp/treemacs-bookmarks.el
@@ -16,8 +16,10 @@
 ;; along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 ;;; Commentary:
-;;; Integrates treemacs with bookmark.el.
-;;; NOTE: This module is lazy-loaded.
+
+;; Integrates treemacs with bookmark.el.
+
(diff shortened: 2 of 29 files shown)


Alexander-Miller:
A surprise to be sure, but a welcome one. Please prefix the commit messages with `[Misc]` in both PRs and we're good to merge.

tarsius:
> A surprise to be sure, but a welcome one.

So far I have been a "outline in the buffer itself" kinda guy. I.e. I make heavy use of `outline-minor-mode`. But this looks like a nice "outline on the side" package. :grinning: 

*Some unsolicited feedback* I noticed that you usually code in windows with lots of vertical space. Not only do you have room for the `treemacs` window (which I struggle with when I display three frames side-by-side) but you additionally seem to have enough room for some very long lines. I would like to encourage you to keep lines below 80 characters, with some exceptions of no more than 95 characters. This helps with that:

```elisp
  (setq-default display-fill-column-indicator-column 80)
  (add-hook 'emacs-lisp-mode-hook 'display-fill-column-indicator-mode)
```

> `[Misc]` in both PRs and we're good to merge.

Done.

Alexander-Miller:
> So far I have been a "outline in the buffer itself" kinda guy. I.e. I make heavy use of outline-minor-mode. But this looks like a nice "outline on the side" package. 

I wouldn't say that tags are one of treemacs' strong points, the file trees themselves are. Unless you like seeing tags of multiple files or prefer treemacs' elisp imenu expression you might be happier with imenu-list.

> I would like to encourage you to keep lines below 80 characters, with some exceptions of no more than 95 characters. 

My laptop is quite small actually, I just never got into the habit of splitting my windows, though that's changing now, so I'll definitely try and shrink things down over time.

> Done.

:+1:

PR #833, opened 2021-08-16
Add org_archive as extension for org files

dakra:
I have almost as many `foo.org_archive` files as `foo.org` files.
It's configurable but since `.org_archive` is the default I guess
it's ok to have it in there.

I also removed the duplicate elisp icon entry in a different
commit (in case you don't want to add org_archive by default).

Thanks,
  Daniel


Alexander-Miller:
The commit messages should be prefixed with `[Icons]`. Other than that LGTM.

dakra:
Thanks. I fixed the commit messages.

Alexander-Miller:
:+1:

PR #845, opened 2021-09-15
[Core] Search for python on Windows even with unexpected errors.

### LINKED ISSUE #844: Python3 Detection May Fail on Windows
If multiple Python installations exist on the machine and one of them happens to be a Cygwin installation using a [magic cookie](https://cygwin.com/cygwin-ug-net/using.html#pathnames-symlinks) symlink, Python3 detection may not succeed.  The reason for this is that the Cygwin installation may use a symbolic link to point at the specific version of Python within the Cygwin installation.  However, that symbolic link may be an old-style link, not a Windows native link. The old-style Cygwin symlinks are just files containing a magic cookie and the link target.

```
C:\>where python
C:\Cygwin\bin\python
C:\Programs\Python37-32\python.exe

C:\>type C:\Cygwin\bin\python
!<symlink>python2.7.exe
C:\>
```

Since the Windows `where` command may list multiple installation paths, it's possible (depending on which is found first in PATH) that the Cygwin version will be found first.  When `treemacs--find-python3` attempts to run the Cygwin path, to obtain the version information, it will fail due to not being able to run this file and `treemacs--find-python3` will report `nil` back to the caller (via the error handler of the `condition-case`), since it aborts searching once any error has occurred.  This prevents subsequent paths which may have been listed in the `where` output from being checked causing treemacs to report back that it couldn't find a Python3 installation when one may in fact have existed.

brownts:
Fixes #844.

Paths to Cygwin binaries utilizing magic cookie symlinks may cause
problems when they show up as a result of running the "where" command.
Keep searching the remaining paths returned from the "where" command
when unexpected errors occur while attempting to obtain the version
from the python binary.

### DIFF
--- a/src/elisp/treemacs-customization.el
+++ b/src/elisp/treemacs-customization.el
@@ -39,12 +39,14 @@
                (s-lines)
                (--first
                 (when (file-exists-p it)
-                  (->> (concat (shell-quote-argument it) " --version")
-                       (shell-command-to-string)
-                       (s-trim)
-                       (s-replace "Python " "")
-                       (s-left 1)
-                       (version<= "3")))))
+                  (condition-case _
+                      (->> (concat (shell-quote-argument it) " --version")
+                           (shell-command-to-string)
+                           (s-trim)
+                           (s-replace "Python " "")
+                           (s-left 1)
+                           (version<= "3"))
+                    (error nil)))))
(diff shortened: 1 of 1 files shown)


Alexander-Miller:
:+1:

PR #866, opened 2021-10-22
Add ability to mark paths for different actions

sdwolfz:
While not technically part of my initial improvement suggestions from #534, I've been struggling with this ever since switching from my last filetree viewer.

The idea is to mark specific paths by navigating to them and pressing a key and once you marked them all, perform the same action on all of them: delete, copy, move, maybe even rename (in the future). Pressing `ESC` (not implemented yet) would technically take you out of the mark selection, and pressing another key would undo this, giving you back your previous selection (reason why I made that `mark-active` variable. Also there should also be (but not yet implemented) a way to visually highlight the marked files, or parent directories in case they're collapsed after marking.

There are may edge cases that I've not handled yet, reason why I haven't exposed the functionality under specific keys.
@Alexander-Miller please take a look at this and leave comments and suggestions. I'll deal with docstrings and the README once we're good. I'd like to introduce this functionality piece by piece, even though it's not complete, and improve on it in future PRs, as it has the potential to be quite big.

In my testing I did:
```elisp
(evil-define-key 'treemacs treemacs-mode-map (kbd "M") #'treemacs-mark-or-unmark-absolute-path-at-point)
(evil-define-key 'treemacs treemacs-mode-map (kbd "N") #'treemacs-reset-mark)
```
Then switched around `N` to the other functions. We'll need to come up with some good keybindings for these as well...

### DIFF
--- a/src/elisp/treemacs-annotations.el
+++ b/src/elisp/treemacs-annotations.el
@@ -23,8 +23,11 @@
 ;;; Code:
 
 (require 'ht)
+(require 'dash)
+(require 'treemacs-async)
 (require 'treemacs-core-utils)
 (require 'treemacs-workspaces)
+(require 'treemacs-async)
 
 (eval-when-compile
   (require 'treemacs-macros)
@@ -37,15 +40,21 @@
 (defconst treemacs--annotation-store (make-hash-table :size 200 :test 'equal))
 (defconst treemacs--annotation-cache (make-hash-table :size 200 :test 'equal))
 
+;; TODO(2022/02/23): clear on file delete
+
 (cl-defstruct (treemacs-annotation
                (:conc-name treemacs-annotation->)
                (:constructor treemacs-annotation->create!)
                (:copier nil))
-  suffix)
+  suffix
+  suffix-value
+  git-face
+  face
+  face-value)
 
 (define-inline treemacs-get-annotation (path)
   "Get annotation data for the given PATH.
-Will return nil if no annotations exist.
+Will return nil if no annotations exists.
 
 PATH: Node Path"
   (declare (side-effect-free t))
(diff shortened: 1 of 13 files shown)


sdwolfz:
>     * I've been reluctant to implement something like this in the past because of how much trouble it would be
>       to keep visual marks alive through file system changes and refreshes etc.
It is much trouble, maybe not implement it initially and not expose the functionality yet, keep it as an "experimental" feature.

>     * I did start building an "annotation" api for use-cases like this on another branch. Only suffix strings for
>       nodes are implemented so far, but having a dedicated feature to use the new annotations will make for a great
>       testing ground. I'll start porting things to your branch when I am done with the current peek-mode I am
>       working on.
Sounds good!

>     * If the marks are properly persistent and visible then I don't think we need explicit tracking 󠁠󠁠à la `mark-active`.
I kept it there more as a reminder that I want a sort of "redo" functionality, like when you select multiple files, press escape to unselect them, then press something else to undo the unselect and have the same selected file set ready for an action. But how this is implemented won't matter.

>     * Marks need to be buffer local, be killed when the buffer is killed/file is deleted etc pp. There'll be some details
>       to track down.
:+1: 

>     * Actions on marks vs actions on the current file should work as in dired. That's the workflow people will be
>       familiar with. In particular that means there shouldn't be any dedicated move/delete-marked-files commands,
>       the default actions should be all dwim-ified.
:+1: , I was also thinking of a "transient state"/"hydra" for this like we have in Spacemacs. What I have in the PR right now was just the quickest POC to get the ball rolling. 

>     * What would rename even do with multiple marked files?
This was a sort of fantasy of mine, renaming multiple files all the same name, and when conflicting, add the `(Copy X)` part in the name like right now, or you could interpret the `%n` or something like this in the name as the placeholder for the counter or something like that. Not sure how useful this is, but I think dired has some multi rename functionality we could get inspiration from. Anyway, not mandatory.

Alexander-Miller:
Small update: what time I've recently had I put into getting the annotation api to work. Integrating that into the current async git-mode setup is a lot more challenging than expected, but I think I've finally found a method that'll work without putting extra work into the critical render loop. I hope I'll be able to push something usable over the weekend. When that's done I'll take a look at your other PR.

Alexander-Miller:
All right, I've updated the the branch with the new api. I'll now take some time to look at the issue that have popped up in the meantime, so there's time for you to look at things as well.

I haven't written a proper guide yet, but for the most part things should be self-explanatory, you won't need more than this:
ann

One thing to keep in mind is that annotation faces come before, and therefore override, git-mode faces. That rule includes directives like `:foreground nil`.

Alexander-Miller:
Stayin alive.

In fact this PR is just about ready now. @sdwolfz Can you give it a try to make sure I didn't miss anything? `treemacs-bulk-file-actions` is a hydra that's the entry point for the whole feature bloc now.

sdwolfz:
@Alexander-Miller I'll see what Ivan do this weekend. 

Thanks for not giving up on me!

sdwolfz:
@Alexander-Miller Tested this out locally and it works amazingly :raised_hands:.

There are some usability improvements I'd like to make later but I'd like to experiment with the functionality for a while before coming up with suggestions. Should be only minor tweaks, non breaking.

Feel free to merge whenever you are comfortable with it. And thank you for all the support!

Alexander-Miller:
Accidentally pushed to master instead of this PR, so we can close here.

PR #867, opened 2021-10-22
Add functionality to create file or directory depending on file path

sdwolfz:
The decision to create either a file or directory is based on the presence of the direcotry separator at the end of the path, similar functionality that I remember having in neotree.

If it were by me I'd remove the other two and only keep this one, but I'm assuming this will be not accepted, so I'm going to add it as a third option like this:

```elisp
(evil-define-key 'treemacs treemacs-mode-map (kbd "ca") #'treemacs-create-file-or-dir)
```

Could also rename the function to `treemacs-create-anything` if that fits the functionality better.

Also, this shares alot of code with `treemacs--create-file/dir`, how would you like this abstracted and code shared? There're a few ways I see this being done and I'd like to hear your opinion first (personally I'd extract the `when-let` part to a different function).

### DIFF
--- a/src/elisp/treemacs-file-management.el
+++ b/src/elisp/treemacs-file-management.el
@@ -274,6 +274,51 @@ IS-FILE?: Bool"
      (treemacs-pulse-on-success
          "Created %s." (propertize path-to-create 'face 'font-lock-string-face)))))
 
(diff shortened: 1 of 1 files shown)


sdwolfz:
@Alexander-Miller any thoughts on this?

PR #868, opened 2021-10-29
[all-the-icons] Changed org-mode icon to the org unicorn in all-the-icons theme

jeslie0:
There is an org icon in all-the-icons and we should use it for the all-the-icons-theme. This change removes org from using the `all-the-icons-octicon "file-text"` icon and makes it now use `all-the-icons-fileicon "org"`, with a light green face.


Alexander-Miller:
LGTM. Prefix the commit message with `[all-the-icons]` and we're good to go.

jeslie0:
Okay, I believe I have done that now!

Alexander-Miller:
You changed the title of the PR in github. It's copied from the commit message, but they are independent things. To change the commit message you have to amend you commit and force push it again.

I took care of that stuff, so we can merge here.

Alexander-Miller:
Aaand I forgot to actually push my changes before pushing the button. Never mind, nothing a little `--force` can't solve.

PR #894, opened 2022-01-22
[Icons] Use elx-light for heex (HTML EEx).

BlindingDark:
[heex](https://hexdocs.pm/phoenix_live_view/assigns-eex.html#:~:text=.heex%20extension%20(HTML%20EEx)) (HTML EEx) is used as a template for Elixir Phoenix.


Alexander-Miller:
:+1:

PR #900, opened 2022-01-29
Assign icons for edn and eln files

dakra:
eln is the new Emacs native compilation extension
edn files contain Clojure syntax


Alexander-Miller:
:+1:

PR #903, opened 2022-01-31
Tab bar mode

### LINKED ISSUE #901: tab-bar-mode scope
Hi, I'm using tab-bar-mode rather than `persp-mode`/`perspective.el` now-a-days for different layouts. It works great for most things. I used to rely on `treemacs-perspective` and its scope. I've started to try writing one for tab-bar, but I'm curious if anyone is in a similar situation and has a solution that works.

### LINKED ISSUE #902: treemacs--select-workspace-by-name behaves differently if there is only one workspace
https://github.com/Alexander-Miller/treemacs/blob/master/src/elisp/treemacs-workspaces.el#L816-L817

When there is only one workspace and you pass a workspace name that does not exist, it will return the first workspace. This differs from when there is 2 or more. In that case, it will return `nil`.

Is that intentional? As far as I can tell it interferes with something like this behaving as expected:

https://github.com/Alexander-Miller/treemacs/blob/master/src/extra/treemacs-perspective.el#L90-L91

aaronjensen:
Adds support for `tab-bar-mode` based on the support for perspectives. From my limited testing so far, it appears to work well and as I'd expect. 

Fixes #901

### DIFF
--- a/README.org
+++ b/README.org
@@ -142,8 +142,9 @@ the current root's parent. Another restriction is that the same part of the file
 part of the workspace. For example it is not possible to have both /Documents and /Documents/ProjectX as projects in the
 same workspace, since internally treemacs heavily relies on every node having a unique natural key in its absolute path.
 Nonetheless the pros certainly outweigh the cons, as a multiroot setup allows to work on multiple projects with any
-combination concern/buffer separating frameworks, be it persp/perspective, eyebrowse, or project.el/projectile. It also
-opens the potential for concurrent display not only of the file system, but e.g. the currently open buffers.
+combination concern/buffer separating frameworks, be it persp/perspective, eyebrowse, tab-bar-mode, or
+project.el/projectile. It also opens the potential for concurrent display not only of the file system, but e.g. the
+currently open buffers.
 
 *** Workspace Selection
 When a workspace is first needed, treemacs will select a workspace in the following manner:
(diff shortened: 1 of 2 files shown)


cmccloud:
Would love to see this get merged.

aaronjensen:
This has been updated to use the fix from #902

README.org
```diff
@@ -509,6 +513,11 @@ discoverability.
     :after (treemacs persp-mode) ;;or perspective vs. persp-mode
     :ensure t
     :config (treemacs-set-scope-type 'Perspectives))
+
+(use-package treemacs-tab-bar ;;treemacs-tab-bar if you use tab-bar-mode
```
Alexander-Miller, line 517: The indentation isn't quite right, it should be 2 spaces further to the right.

README.org
```diff
@@ -997,7 +1006,7 @@ into their own module in ~treemacs-icons.el~).
 If you want to delve into the treemacs' code base, check out [[https://github.com/Alexander-Miller/treemacs/wiki][the wiki]] for some general pointers.
 
 * Dependencies
- - emacs >= 26.1
+ - emacs >= 26.1 (>= 27.1 for tab-bar)
```
Alexander-Miller, line 1009: New module will need a new tag line for commit messages. Let's go with `[Tab Bar]` on this one.

Alexander-Miller (changes requested):

aaronjensen:
@Alexander-Miller thank you, updated.

Alexander-Miller (approved):

Alexander-Miller:
Next stop: melpa https://github.com/melpa/melpa/pull/7913

PR #915, opened 2022-03-07
[Workspaces] Use project-root function.

bastianbeischer:
Fixes compatibility with emacs trunk, where an additional item was added to the
project-current list.


Alexander-Miller:
:+1:

Issue #921, opened 2022-03-14
Emacs locks up when treemacs is run in git repositories with large ignored trees


BeneSim:
I know the title is a mouthful, so let me quickly explain what i mean with this.
I regularly work with git projects (OpenEmbedded) that have an extremely large tree that gets ignored by .gitignore. For example in a regular OpenEmbedded based project I'll ignore everything in the `build` directory except the `conf` folder. Since OpenEmbedded effectively builds an entire Linux Distribution from scratch (including the required compilers) the ignored tree is massive.

For example: Issuing the *default* `git status --porcelain --ignored` which results in `git status --porcelain --ignored=traditional` on such a projects leads to 

```
Executed in  287.46 secs    fish           external
   usr time   17.26 secs  975.00 micros   17.26 secs
   sys time   49.40 secs  144.00 micros   49.40 secs
```

Using `git status --porcelain --ignored=matching` instead on the same projects leads to

```
Executed in  106.59 millis    fish           external
   usr time   65.52 millis    0.00 millis   65.52 millis
   sys time   67.24 millis    1.36 millis   65.88 millis
```

You can't get all ignored files via `--ignored=matching --untracked-files=all` since - in my understanding - `--ignored=matching` exits early, i.e. if the root node matches the ignore pattern everything under the root node is skipped. This is only possible with `--ignored[=traditional]`, however, even when you don't use `--untracked-files=all` (which treemacs doesn't) git seems to visit all files recursively.

I haven't dug too deep into the internals of git and the documentation isn't that clear to me. The only thing I can tell for sure - at least for me - is that both options lead to exactly the same output but with a couple of orders of magnitude more performance in the case of `--ignored=matching`.

I'll also issue a PR that uses `--ignored=matching` everywhere.

PR #922, opened 2022-03-14
Use `--ignored=matching` with `git status`

BeneSim:
The default `--ignored[=traditional]` seems to recursively visit all
files within a project even if `--untracked-files=all` is not
supplied. This may take a lot of time for project with a very large
tree where the root node is ignored.

See #921 for a more thorough explanation.


Alexander-Miller:
:+1:

PR #923, opened 2022-03-14
clarify side effects of treemacs-display-current-project-exclusively

augustfengd:
Hi!

I feel like the original wording is not too clear about the fact that the projects would be actually wiped from the workspace file. I had thought that the command would just temporarily focus on a single project.


Alexander-Miller:
:+1:

PR #926, opened 2022-03-18
Removing custom Gradle icon

bricka:
Support for .gradle files was added to all-the-icons in this MR:
https://github.com/domtronn/all-the-icons.el/pull/311


Alexander-Miller:
:+1:

PR #939, opened 2022-05-15
Fix a few custom types

tarsius:


src/elisp/treemacs-customization.el
```diff
@@ -171,7 +171,8 @@ The choices are
    level
  - `block' to use a thick '██' block interspersed at every second indentation
    level"
-  :type '(choice (const :tag "Line" 'line) (const :tag "Block" 'block))
+  :type '(choice (const :tag "Line" line)
+		 (const :tag "Block" block))
```
Alexander-Miller, line 175: Looks like you didn't check the box that'll let me push to your branch, so now you get to make fixes yourself :man_shrugging: 

The indentation's wrong here.  The commit message should also start with `[Misc]`.

Alexander-Miller (changes requested):

src/elisp/treemacs-customization.el
tarsius: > Looks like you didn't check the box that'll let me push to your branch, so now you get to make fixes yourself man_shrugging

That's weird. I used Forge to create the pull-request and that is supposed to always enable this. I'll have to check if there is a regression.

> The indentation's wrong here.

And I even briefly considered turning on `whitespace-mode` to check, but decided that you surely have enforced your preferred indentation style. :roll_eyes:

I prefer spaces too but I have left the global value of `indent-tabs-mode` alone because that forces me to set it in the `.dir-locals.el` file of all my repositories, and that really is the only way to avoid the occasional wrong indentation in a pull-request because a contributor didn't bother to double check. I recommend you do the same. :grinning:

tarsius: > That's weird. I used Forge to create the pull-request and that is supposed to always enable this. I'll have to check if there is a regression.

I have used the API to check the value. It claims that `maintainerCanModify` is true for this pull-request.

Alexander-Miller (approved):

PR #957, opened 2022-06-27
Fix README Typo

christmoore:
Noticed this on a fresh install of treemacs

Causes `Error (use-package): treemacs/:config: Wrong type argument: symbolp, 'treemacs-hide-dot-git-directory`


CaioVdmm:
I can confirm thats an error in the progn part cause treemacs worked like a charm after i deleted the progn part of the default config.
maybe progn think argument is wrong even it's been correct
sorry for my english

Alexander-Miller:
:+1:

PR #960, opened 2022-07-08
Fix visit-node-ace not focusing when there are multiple emacs frames.

basaran:
- Changed to (ace-window)

When you have multiple emacs frames, and have

```
(setq treemacs-default-visit-action 'treemacs-visit-node-ace)
```
It wouldn't raise the other frame through X.

### DIFF
--- a/src/elisp/treemacs-interface.el
+++ b/src/elisp/treemacs-interface.el
@@ -40,7 +40,9 @@
   (require 'treemacs-macros))
 
 (autoload 'ansi-color-apply-on-region "ansi-color")
-(autoload 'aw-select "ace-window")
+
+(treemacs-import-functions-from "ace-window"
+  ace-select-window)
 
 (treemacs-import-functions-from "cfrs"
   cfrs-read)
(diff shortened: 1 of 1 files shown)


Alexander-Miller:
Sorry for the late response, I've only now finally got deep enough into my backlog to look at your PR. Review's coming in in a bit.

src/elisp/treemacs-interface.el
```diff
@@ -251,7 +251,7 @@ Stay in the current window with a single prefix argument ARG, or close the
 treemacs window with a double prefix argument."
   (interactive "P")
   (treemacs--execute-button-action
-   :window (aw-select "Select window")
```
Alexander-Miller, line 254: The autoload for `aw-select` in line 43 can be removed as well.

src/elisp/treemacs-interface.el
```diff
@@ -251,7 +251,7 @@ Stay in the current window with a single prefix argument ARG, or close the
...
+   :window (ace-window "Select window")
```
Alexander-Miller, line 254: `ace-window` just calls `ace-select-window`, just without prefix arg options we don't need. Might as well use that.

src/elisp/treemacs-interface.el
Alexander-Miller: Instead we need this line:

```
(treemacs-import-functions-from "ace-window"
  ace-select-window)
```

src/elisp/treemacs-interface.el
Alexander-Miller: This commit message should be prefixed with `[Interface]`.

Alexander-Miller (changes requested):

Alexander-Miller (approved):

PR #971, opened 2022-08-24
Prefer `display-buffer' to home-grown window management

mickeynp:
This PR replaces internal calls made directly to both `switch-to-buffer`, `display-buffer-in-side-window`, et al. with the preferred method of window management in *newer* Emacsen.

Instead of setting window dedication, window size, and side bar position directly, the code defers this to `display-buffer` with a suitable `ACTION` strategy.

Now a user can override how Treemacs displays its main window by configuring `display-buffer-alist`:

```elisp
("^ \\*Treemacs" display-buffer-in-side-window
  (side . right)
  (window-width . 30))
```

Note that although I've tested it, I am not familiar with the complications of the code writ large, so it's possible there are some feature regressions. But it works well, and this is guaranteed to place nice with all and sundry that use `display-buffer-alist`.

### DIFF
--- a/src/elisp/treemacs-core-utils.el
+++ b/src/elisp/treemacs-core-utils.el
@@ -1039,16 +1039,21 @@ Will be added to `treemacs-ignored-file-predicates' on Macs."
 
 (defun treemacs--popup-window ()
   "Pop up a side window and buffer for treemacs."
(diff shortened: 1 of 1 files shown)


Alexander-Miller:
:+1:

PR #983, opened 2022-10-10
[Interface] fix copy-functions used `kill-new` not work in emacs-29

sunlin7:
There're an error message when try `treemacs-copy-relative-path-at-point` with emacs29, 
> Wrong type argument: stringp, nil

The reason is the `kill-new` function will return the string result in emacs28, but it return nil in emacs29.
And the manual for `kill-new` didn't promise to return the target string.
This PR will fix the error message in emacs29.

Please help review the change. Thanks.


Alexander-Miller:
:+1:

sunlin7:
@Alexander-Miller Thanks !

Issue #956, opened 2022-06-23
Python traceback appearing randomly(?) in minibuffer


montaropdf:
Hello,

Since the last package update I perform, I see the following python traceback appearing in the minibuffer. I have not identified a pattern to reproduce the error.

```
[Treemacs] Update of node "/home/smals-re/repos/gcloudadm/src/gcloudadm/report/utils.py" failed with status "exited abnormally with code 1" and result
[Treemacs] "Traceback (most recent call last):
  File "/home/smals-re/.emacs.d/elpa/treemacs-20220622.801/treemacs-single-file-git-status.py", line 108, in <module>
    main()
  File "/home/smals-re/.emacs.d/elpa/treemacs-20220622.801/treemacs-single-file-git-status.py", line 76, in main
    elif (changed_proc.wait() == 0 and changed_proc.stdout.read1() != b''):
TypeError: read1() takes exactly one argument (0 given)"
Wrote /home/smals-re/.emacs.d/.bookmarks.~undo-tree~
```

Alexander-Miller:
What python version are you on - specifically the value of `treemacs-python-executable`?

What's the output of `python /treemacs-install-location/treemacs-single-file-git-status.py /your/file X`?

What's the output if you change the script to use `read()` instead of `read1()`?

montaropdf:
Sorry for the late reply.

I finally find a way to prevent those messages.

By default, it seams that python 3.6 is used on my system. So I forced emacs and treemacs to use python 3.8. Since then I don't see anymore those errors.

Variable modified to force the use of python 3.8.
`    (setq treemacs-python-executable "/usr/bin/python3.8")`

PR #985, opened 2022-10-12
Support Python 3.6 in treemacs-single-file-git-status.py

jrosdahl:
Thanks for a nice package!

I'm stuck on a system with Python 3.6 and got the problem described in #956.

The issue is that [io.BufferedReader.read1](https://docs.python.org/3.6/library/io.html#io.BufferedReader) needs an argument in Python 3.6, and `-1` is not supported. But if I understand the code correctly, it should be enough to read one byte, so I made it so locally and it seems to work fine for me. Perhaps the fix could be acceptable for treemacs master as well?


Alexander-Miller:
:+1:

PR #989, opened 2022-11-01
Enable evil-escape integration if a newer version is used

smile13241324:
I am trying to revive evil-escape, as part of this I have integrated most of the PRs from the original repository and did fix the treemacs integration. The package is not yet available on Melpa but is already used by spacemacs by means of the github fetcher.

This PR updates the compatibility layer for treemacs to test for a new version variable and only disable the integration if the old version is found aka the variable is not defined, this would enable users who have already switched to the new fork to use the full feature set.


Alexander-Miller:
:+1:

Issue #1002, opened 2023-02-17
Handle persp rename


samrjack:
## Summary
I use persp quite a bit and keep running into issues with treemacs sometimes breaking after switching projects and showing incorrect projects or the correct projects but not letting me modify them etc. For example, if there is more than one project in a the treemacs workspace, trying to delete one may lead to the error message `Cannot delete the last project.`

## Proposed root cause
I believe this comes from the fact that treemacs does not handle persp's rename functionality (defined in persp.el [here](https://github.com/Bad-ptr/persp-mode.el/blob/master/persp-mode.el#L2860) and in perspective.el [here](https://github.com/nex3/perspective-el/blob/master/perspective.el#L1156)). Instead, from what I can see, treemacs keeps the old perspective name as the current treemacs workspace name. Then, after changing perspectives to something else and switching back again, the current treemacs workspace is named after the current perspective and is missing projects.

## Proposed solution
Based on the above, I believe the easiest path forward is hook into persp.el (and perspective.el) rename hooks and rename the current treemacs workspace to reflect that changed perspective name.

Persp.el provides the hook var `persp-renamed-functions` and perspective.el provides the hook var `persp-after-rename-hook` so there would need to be some logic to work with both but it should be doable.

## Unknowns
I am happy to go in myself to add this, but here are the areas I'm grey on as far as side-effects are concerned.
- Since using perspective causes lots of treemacs workspaces to be created, I don't know how trying to use a name that already exists would affect the perspective.
- I don't know if there are other variables that would need to be touched to maintain internal consistency within treemacs.
- I believe this would require creating a more generic rename method since the only one right now (that I can find) requires user interaction.

If anyone has any insights into any of the discussion above, please let me know.

PR #1004, opened 2023-02-22
[Persp] Rename treemacs workspace when persp is renamed

samrjack:
## Summary
My attempt to fix [issue 1002](https://github.com/Alexander-Miller/treemacs/issues/1002). I added logic to programmatically rename a treemacs workspace and then added hooks for persp-mode and perspective.el.

## Callouts
- I've only tested the `treemacs-perspective--on-perspective-rename` function manually since I don't use `perspective.el`.
- There are more places where the rename function could be inserted to be used, but I left them alone for the time being as I don't know the treemacs codebase that well.
- Along with the previous point, I don't know if the rename function was written as efficiently as possible, so let me know if there are obvious improvements to be made.


Alexander-Miller:
> samrjack closed this 2 days ago

Huh?

> Files changed (254)

What happened here?

samrjack:
> Huh?

The pull request doesn't seem to like when I push changes to a commit so it closes the PR. Apologies for the mess as I've not seen this behavior in other repos.

> > Files changed (254)
> What happened here?

Somehow my fork condensed all the history down into a single commit 😅 I fixed it up so that the changes are in clean commits as originally intended. The correct diff can be seen [here](https://github.com/Alexander-Miller/treemacs/compare/master...samrjack:treemacs:master). I'll open the merge request for it once I figure out why it kept closing itself so I can prevent a recurrence of this behavior.

PR #1006, opened 2023-03-01
[Persp] Rename treemacs workspace when persp is renamed

samrjack:
## Summary
My attempt to fix [issue 1002](https://github.com/Alexander-Miller/treemacs/issues/1002). I added logic to programmatically rename a treemacs workspace and then added hooks for persp-mode and perspective.el.

I also did some spelling fixes in other files. All the spelling fixes are in a separate commit for logical clarity.

## Callouts
- I've only tested the `treemacs-perspective--on-perspective-rename` function manually since I don't use `perspective.el`.
- There are more places where the rename function could be inserted to be used, but I left them alone for the time being as I don't know the treemacs codebase that well.
- Along with the previous point, I don't know if the rename function was written as efficiently as possible, so let me know if there are obvious improvements to be made.

## Other
Apologies for the multiple pull requests. I won't go into details about why that happened, but this one is finally in a state where it's showing the correct set of changes.

### DIFF
--- a/src/elisp/treemacs-workspaces.el
+++ b/src/elisp/treemacs-workspaces.el
@@ -662,8 +662,12 @@ Return values may be as follows:
      (treemacs-return
       `(success ,new-workspace)))))
 
-(defun treemacs-do-rename-workspace ()
+(defun treemacs-do-rename-workspace (&optional workspace new-name)
   "Rename a workspace.
+
+Takes either a WORKSPACE and a NEW-NAME as arguments or reads them
+interactively.
+
 Return values may be as follows:
 
 * If the given name is invalid:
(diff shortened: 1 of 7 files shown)


Alexander-Miller:
Thanks for the well documented contribution. I've made a few adjustments to better integrate your changes with the rest of treemacs. Please try it out and let me know if things now work as they should.

If all goes well also on your end I'll go ahead and merge.

samrjack:
Awesome, I'll take a look this weekend and get back to you.

samrjack:
I tried it out and it seems to work well. Thanks for patching it up!

Alexander-Miller:
:+1:

PR #1015, opened 2023-04-10
Add alphabetically and numerically sorting options

hwiorn:
Treemacs doesn't support numerically sorting currently.

I add feature using `string-version-lessp`


Alexander-Miller:
Thanks :+1: 
(FYI I removed the part where the default sorting is changed, otherwise this feature is merged now)

Issue #1017, opened 2023-04-13
Invalid image type 'svg'


Randy1Burrell:
**What's the problem?**
Recently after Emacs elpa upgraded the packages I am getting the error "Invalid image type 'svg'" from Treemacs.

**Version**
 Emacs 28.2

**OS**
MacOS

**What do you expect to happen?**
I expect Treemacs to work.

chrishowejones:
I get the same issue.

Alexander-Miller:
I have recently added a few icons in the svg format because it has a few advantages over the pngs currently in use - svgs can be upscaled without pixelation which is very useful for people with large 4k displays, they also have no trouble with overlays changing their background color, which is something I need to fix manually with png files.

For all intents and purposes svg files are the future of treemacs icons, it just will take a long time to migrate the 300 or so icons that already exist. I just hope lack for svg support is not a common issue, or at least it's easy to rectify.

Do you also see the error when you try this?

`(insert (propertize " " 'display '(image :type svg :file "treemacs install dir/icons/default/vsc/dir-git-open.svg")))`

What does `(image-type-available-p 'svg)` return?

> I get the same issue.

Are you also on a mac?

My google-fu says Emacs has supported svgs since 2007, there's 7 year old threads asking how to compile Emacs with svg support like [this](https://www.reddit.com/r/emacs/comments/33zur0/how_do_i_enable_svg_support_for_emacs_on_os_x/).  Maybe you just need to install the right version?

chrishowejones:
@Alexander-Miller I am on Ubuntu 22.10 and Emacs 29  `Emacs 29.0.50 (build 2, x86_64-pc-linux-gnu, GTK+ Version 3.24.34, cairo version 1.16.0)` built from github source some months ago.

`(insert (propertize " " 'display '(image :type svg :file "treemacs install dir/icons/default/vsc/dir-git-open.svg")))` returns `nil`

`(image-type-available-p 'svg)` also returns `nil`

Randy1Burrell:
I'm using the native packages built for Mac. What do your think the solution is here?

Alexander-Miller:
> I'm using the native packages built for Mac. What do your think the solution is here?

You might be affected by [this issue](https://debbugs.gnu.org/cgi/bugreport.cgi?bug=59081). That means you should:

- Wait until I remove svg icons if your Emacs doesn't support them. The fix will be my current priority.
- Get yourself an Emacs that does support them if you want to see the new directory icons.
[...]

PR #1020, opened 2023-04-20
Handle unsupported images.

Xaldew:
The recent update to use SVG images appear to be quite fragile, breaking a number of common use-cases including TUI usage that should have been filtered out elsewhere.

This change makes sure that if the image-type is not supported, the fallback is directly used.

This should solve issues encountered in #1017.


Xaldew:
This has been fixed in another patch quite a while ago at this point.

PR #1037, opened 2023-07-06
[all-the-icons] Update to all-the-icons 6.0 (SVG)

krn-robin:
Migrated all-the-icons support from 5.x to the upcoming 6.0 (SVG) release.

Example setup for testing (needs all-the-icons 6.0 from git):
```
;; All the icons
(use-package all-the-icons
  :if (display-graphic-p)
  :vc (:url "https://github.com/domtronn/all-the-icons.el.git"
	    :branch "svg"
	    :rev :newest))

;; Treemacs all-the-icons integration
(use-package treemacs-all-the-icons
  :vc (:url "https://github.com/krn-robin/treemacs.git"
	    :branch "all-the-icons6"
	    :lisp-dir "src/extra"
	    :rev :newest)
  :after (treemacs all-the-icons)
  :config
  (treemacs-load-theme "all-the-icons"))
```

Pull request marked as draft for now, since all-the-icons 6 has not been released yet.


Alexander-Miller:
So on the one hand I've got this PR that has the potential to let me outsource all my icon needs and save hundreds of lines of some of the more complicated code I've written. On the other hand I've got issues of people who have all sorts of troubles with the few svg icons that I have introduced myself. Why can't thing things ever be simple?

Anyway, thanks for the PR :+1:.  Please let me know when v6 is out.

webdevred:
While I think this would be a good a addition for the package I wonder, with this added, will it still be possible to not have any icons at all like I do now?
I am kind of a minimalist myself so I would prefer to not have any icons or any packages related to icons installed.

Alexander-Miller:
It's not an addition, it's updating the already pre-existing theme to use the newest all-the-icons api, so I'm not sure how exactly this would affect your icon-less config.

If it does then the easiest way to get rid of icons is by creating your own theme that defines nothing except the bare basics like fallbacks.

webdevred:
@Alexander-Miller Thanks for your reply! Could you also please at my issue regarding Treelib and extensions please?

wyuenho:
Just saw this! Thanks for making this PR. FYI, the SVG branch is still in progress. There are a few issues that I aim to fix but haven't had the time. Hopefully this coming holiday will give me some time!

PR #1040, opened 2023-07-24
Fix treemacs-delete-other-windows

mpedramfar:
Hi and thanks for the great package!

Currently `treemacs-delete-other-windows` calls `treemacs--select-not-visible-window`, with the assumption that the treemacs window is not visible anymore. The assumption can be false since `delete-window` doesn't delete treemacs in all setups (e.g. when `treemacs-display-in-side-window` is `t` and treemacs is in a side window). The result is that calling `treemacs-delete-other-windows` on the treemacs window splits the treemacs side window vertically. This PR fixes that.


Alexander-Miller:
:+1:

PR #1066, opened 2023-09-30
[Treelib] Let extensions control help-echo

kurnevsky:
The same as was done for extensions in 0a3d2bd67cfb56468fbeea1e0be86436c049e931


kurnevsky:
@Alexander-Miller could you take a look please? :)

Alexander-Miller:
:+1:

PR #1083, opened 2024-01-04
Strip filename after splitting a line

auxsvr:
Prevents incorrect quoting of filenames.

Resolves https://github.com/Alexander-Miller/treemacs/issues/1084.


Alexander-Miller:
:+1: 

(I wanted to correct the commit message, but didn't seem to have permission to push to your fork, so I cherry-picked the commit myself instead).

PR #1103, opened 2024-04-30
[Mouse] Enable/disable moving files by dragging

sebastiaanspeck:


src/elisp/treemacs-mouse-interface.el
```diff
@@ -37,6 +37,16 @@
...
   (require 'treemacs-macros))
 
+(defgroup treemacs-mouse nil
+  "Treemacs+mouse configuration options."
+  :group 'treemacs-mouse
+  :prefix "treemacs-mouse-")
+
+(defcustom treemacs-mouse-move-files-by-dragging t
+  "If non-nil, move files by dragging with your mouse inside treemacs."
+  :group 'treemacs-mouse
```
Alexander-Miller, line 47: The new variable needs to be advertised in 2 places in the readme: in the install use-package example and the table of all configuration variables. There should also be a line in the Changelog.org.

Alexander-Miller (changes requested):

src/elisp/treemacs-customization.el
```diff
@@ -740,6 +746,11 @@ fallback."
   :type 'boolean
   :group 'treemacs-follow)
 
+(defcustom treemacs-move-files-by-mouse-dragging t
+  "When non-nil `treemacs--drag-move-files' will move files by dragging with your mouse inside treemacs."
```
Alexander-Miller, line 750: This line needs to be less than 80 characters so the build will pass.

Your commits should also be squashed when you're done.

Alexander-Miller (changes requested):

Alexander-Miller (approved):

Alexander-Miller:
:+1:

Issue #879, opened 2021-11-26
Default evil cursor + fringe looks like a bug


aaronjensen:
<img width="221" alt="CleanShot 2021-11-26 at 11 52 43@2x" src="https://user-images.githubusercontent.com/8588/143612086-6bd1258a-c0a1-4835-a9d7-4bbaf098e327.png">

I spent some time trying to figure out why a weird L was rendering. It looks like this is what is expected though if I'm using evil since it forces it to use an hbar? Is there no way to hide the cursor w/ evil?

Alexander-Miller:
Treemacs' default is to show no cursor whatsoever. If you're seeing it you either set `treemacs-show-cursor` to t or you are not using the treemacs-evil package.

Your theme also looks strange - if that black and bold color comes from git-mode it means that `hl-line-mode` is invisible to you and you have only the fringe indicator to tell you which line you're in.

aaronjensen:
```el
(use-package treemacs-evil
  :after (treemacs evil)
  :demand t)
```

<img width="679" alt="CleanShot 2021-12-01 at 22 05 40@2x" src="https://user-images.githubusercontent.com/8588/144350415-94958295-6f85-4176-95b6-d0ae6e36591b.png">

<img width="664" alt="CleanShot 2021-12-01 at 22 25 44@2x" src="https://user-images.githubusercontent.com/8588/144352391-5eba3070-1177-4276-a69e-b594d9ba82a9.png">

The cursor being shown is a hbar, rather than bar. I have no idea what's changing it to hbar since it's clearly bar when the state is created. Maybe it's something else in my config. I'll see if I can track that down.

I do have a visible hl-line btw, it's just very faint and may be hard to see on some monitors.

aaronjensen:
Oh. It's treemacs-mode.el itself: https://github.com/Alexander-Miller/treemacs/blob/d00cc19cab8df4fec7ed6608e00bd16fe797369a/src/elisp/treemacs-mode.el#L274

I assume that's actually supposed to be `bar`?

aaronjensen:
https://github.com/Alexander-Miller/treemacs/pull/881

Alexander-Miller:
Are you sure your PR solves your problem? It shouldn't make a difference whether the cursor is a bar or an hbar, that zero determines the height, and a height of zero means that the cursor is not visible. A simple experiment with `(setq-local cursor-type '(hbar . 0))` confirms that, at least on my end. Does your Emacs work differently?

[...]

PR #1114, opened 2024-06-29
Cursor type is set to nil in evil when treemacs-show-cursor is disabled

aaronjensen:
Proper fix for #879 

Follow up for #881


Alexander-Miller:
:+1:

PR #1122, opened 2024-10-02
[Tab Bar] Correct method for getting current tab

aaronjensen:
If a tab's selected buffer is the Treemacs buffer when switching away from a tab and back to that tab, the Treemacs buffer will temporarily not be considered part in the current tab. My original technique for getting the current tab was either always wrong or based on the best we could do at the time with the APIs available. This method of getting the current tab seems to work consistently.


aaronjensen:
This is still necessary to correct a mistake.

Alexander-Miller (approved):

Alexander-Miller:
:+1:

Issue #1123, opened 2024-10-03
treemacs icons shouldn't change 'modifed flag in regular dired buffers.


agzam:
When icons are enabled in Dired - `(add-hook 'dired-mode-hook #'treemacs-icons-dired-mode)`, the buffer becomes "unsaved" - attempts to kill a Dired buffer cause a popup to appear, asking if you want to save it. It looks like the `treemacs-icons-dired--display` hooked onto `'dired-after-readin-hook` cause that and it shouldn't. Specifically it calls `treemacs-icons-dired--display-icons-for-subdir` which at some point does `(insert icon)`, and that causes the buffer to become "modified".
Dired buffers shouldn't be "modified".

PR #1124, opened 2024-10-03
Dired buffer's modifed flag should not change

agzam:
fixes https://github.com/Alexander-Miller/treemacs/issues/1123


Alexander-Miller:
:+1:

PR #1129, opened 2024-11-13
add: Icon for ClojureDart filetype

niontrix:
Why?:
- ClojureDart is a new and coming dialect of Clojure. Thus its files should have the same icon as other Clojure files.

This change addresses the need by:
- Add "cljd" as additional Clojure file extension


Alexander-Miller:
:+1:

PR #1192, opened 2026-05-16
[Visit] Skip side and dedicated windows when picking next window

rain-64:
## The bug

`treemacs-visit-node-no-split` is documented as opening a node ["in the window next to treemacs, the exact selection is determined by `next-window`"][docstring]. But `next-window` does not filter side or dedicated windows, so when treemacs sits next to one - e.g. a strongly-dedicated sidebar buffer in a three-window layout `[sidebuf | main | treemacs]` - the visit action selects that side window first. Since the target cannot host the new buffer, `find-file` falls back to `display-buffer`, which splits a different window. You end up with four windows where you expected to reuse one of the three.

The same root cause affects `treemacs-visit-node-vertical-split` and `treemacs-visit-node-horizontal-split`: side windows cannot be split, so picking one is always wrong.

### Reproduction

This snippet shows the underlying `find-file` behavior on a strongly-dedicated side window:

```elisp
emacs --batch --eval "
(progn
  (let* ((main-buf (get-buffer-create \"*main*\"))
         (left-buf (get-buffer-create \"*left*\"))
         (right-buf (get-buffer-create \"*right*\")))
    (set-window-buffer (selected-window) main-buf)
    (let ((left-win (display-buffer-in-side-window left-buf '((side . left) (slot . 0))))
          (right-win (display-buffer-in-side-window right-buf '((side . right) (slot . 0)))))
      (set-window-dedicated-p left-win t)
      (select-window left-win)
      (find-file \"/tmp/foo.txt\")
      (princ (format \"windows: %s\n\" (mapcar (lambda (w) (buffer-name (window-buffer w))) (window-list))))))"
```

Today it reports four windows. The same sequence triggered by `o o` from a treemacs buffer on the right side of that layout produces the same four-window result.

## The fix

Add a small helper, `treemacs--next-non-side-window`, that cycles through the frame's windows in `next-window` order and skips any window that is a side window (`window-side` parameter) or dedicated (`window-dedicated-p`). Route the `treemacs--execute-button-action` macro's fallback through it.

```elisp
(defun treemacs--next-non-side-window (from)
  "Find the next window after FROM that is not a side or dedicated window.
Cycles through the frame's windows in `next-window' order.  Falls back to
the immediate next window if every candidate is a side/dedicated window."
  (let* ((first-next (next-window from nil nil))
         (candidate first-next))
    (catch 'found
      (while t
        (unless (or (window-parameter candidate 'window-side)
                    (window-dedicated-p candidate))
          (throw 'found candidate))
        (setq candidate (next-window candidate nil nil))
        (when (eq candidate from)
          (throw 'found first-next))))))
```

The macro changes from:

```elisp
(select-window (or ,window (next-window (selected-window) nil nil)))
```

to:

```elisp
(select-window (or ,window (treemacs--next-non-side-window (selected-window))))
```

The explicit `:window` argument used by callers (e.g. `get-mru-window`, `ace-select-window`, an existing window already showing the file) is still respected exactly. Only the fallback path changes.

## Edge cases

- **Three windows, treemacs on the right, dedicated side buffer on the left**: `o o` opens in the main window (the reproduction above). Fixed.
- **Two windows, treemacs + one normal window**: behavior unchanged - opens in that one window.
- **Treemacs is the only window**: the macro's existing `ensure-window-split` branch handles this; the helper is not reached.
- **Pathological case where every other window is dedicated/side**: helper falls back to the original `next-window` result, so behavior is no worse than today.
- **Explicit `:window` argument**: short-circuits via `or`, helper not called.

## Functions that benefit

Every caller of `treemacs--execute-button-action` that does not pass an explicit `:window` - most prominently the visit actions documented as "next to treemacs":

- `treemacs-visit-node-no-split`
- `treemacs-visit-node-vertical-split`
- `treemacs-visit-node-horizontal-split`

Callers that already pass `:window` explicitly (MRU, ace-window, reuse-existing-window) are unaffected because the explicit argument short-circuits the fallback.

## Tests

Added three buttercup specs under `treemacs--next-non-side-window` in `test/treemacs-test.el` covering the three meaningful branches (skip-dedicated-side, two-normal-windows, all-dedicated fallback). The full suite (`make test`) is green at 257 / 257.

[docstring]: src/elisp/treemacs-interface.el


rain-64:
Superseded by #1194. That PR is a strict superset — same dedicated/side-window fix this one introduces, plus a directional fix that handles the case where treemacs sits on the right of multiple normal windows (where `next-window` wraps to the leftmost rather than picking the spatially adjacent one), plus docstring updates to drop the now-stale `next-window` references from `treemacs-visit-node-no-split`, `-vertical-split`, and `-horizontal-split`. Single coherent change instead of stacked PRs.

PR #1193, opened 2026-05-16
[Visit] Pick the window spatially adjacent to treemacs

rain-64:
**Stacked on top of [#1192](https://github.com/Alexander-Miller/treemacs/pull/1192).** This PR builds directly on the "skip side and dedicated windows" fix and assumes that helper exists. Merging this one absorbs both fixes; merging only #1192 leaves this issue open.

## The bug

`treemacs-visit-node-no-split` is documented as opening a node ["in the window next to treemacs"][docstring]. The macro's fallback path uses `(next-window (selected-window) nil nil)`, which cycles through the frame in Emacs's window-list order - typically left-to-right, top-to-bottom, wrapping at the end. In a layout where treemacs is on the right of two normal windows:

```
[ *A* | *B* | treemacs ]
```

`(next-window treemacs-window)` returns `*A*` (the leftmost window, wrapping around), not `*B*` (the one spatially adjacent to treemacs). The previous fix in #1192 only filtered out side and dedicated windows from the wrap-around result; since neither `*A*` nor `*B*` is a side or dedicated window, the fix had no effect on this case.

This means `o o` in the layout above opens the file in `*A*`, jumping a window over the one the user expected.

### Reproduction (batch mode)

```elisp
(progn
  (require 'treemacs-core-utils)
  (let* ((a (get-buffer-create "*A*"))
         (b (get-buffer-create "*B*"))
         (tm (get-buffer-create "*tm*")))
    (set-window-buffer (selected-window) a)
    (let* ((b-win (split-window-right))
           (t-win (split-window b-win nil 'right)))
      (set-window-buffer b-win b)
      (set-window-buffer t-win tm)
      (message "next-window from treemacs: %s"
               (buffer-name (window-buffer (next-window t-win nil nil))))
      (message "window-in-direction 'left from treemacs: %s"
               (buffer-name (window-buffer (window-in-direction 'left t-win)))))))
```

Output (master, even after #1192):
```
next-window from treemacs:                       *A*    ← what the macro picks
window-in-direction 'left from treemacs:          *B*    ← what the user expects
```

## The fix

Add `treemacs--window-next-to-treemacs`, which derives a direction from `treemacs-position` (`'left` when treemacs is on the right, `'right` when on the left) and asks Emacs directly via `window-in-direction`. If that returns a sensible window (non-nil, not a side window, not dedicated) the helper returns it; otherwise it falls back to `treemacs--next-non-side-window` from #1192:

```elisp
(defun treemacs--window-next-to-treemacs (from)
  "Find the window spatially adjacent to FROM (the treemacs window).
Uses `window-in-direction' to pick the window immediately next to
treemacs, based on `treemacs-position' (the opposite direction is
where the rest of the frame lives).  Skips side and dedicated windows.
Falls back to `treemacs--next-non-side-window' when no suitable
adjacent window exists."
  (let* ((direction (if (eq treemacs-position 'right) 'left 'right))
         (adjacent (window-in-direction direction from)))
    (if (and adjacent
             (not (window-parameter adjacent 'window-side))
             (not (window-dedicated-p adjacent)))
        adjacent
      (treemacs--next-non-side-window from))))
```

The macro change is a one-line swap:

```diff
-(select-window (or ,window (treemacs--next-non-side-window (selected-window))))
+(select-window (or ,window (treemacs--window-next-to-treemacs (selected-window))))
```

The explicit `:window` argument used by callers (`get-mru-window`, `ace-select-window`, existing window showing the file) still short-circuits via `or` and is unaffected.

## Edge cases

- **`[A | B | treemacs]`, treemacs on right**: opens in `B` (adjacent), no longer `A`.
- **`[treemacs | B | C]`, treemacs on left**: opens in `B` (adjacent). Previously also `B` here, since `next-window` and `window-in-direction` agree when there's no wrap-around; behavior unchanged.
- **Two windows, treemacs + one normal**: `window-in-direction` returns that one window. Behavior unchanged.
- **Treemacs is the only window**: `window-in-direction` returns `nil`; helper falls through to `treemacs--next-non-side-window`, which in turn is short-circuited by the macro's existing `ensure-window-split` branch. No regression.
- **Adjacent slot is a side or dedicated window**: helper rejects it and falls through to the cycle-and-skip helper from #1192. Behavior matches the prior fix.
- **`treemacs-position` is `'left` (the default)**: direction is `'right`; in a two-window layout `[treemacs | X]` this returns `X`, identical to the prior behavior.
- **Explicit `:window` arg**: short-circuited by `or`, helper not reached.

## Functions that benefit

The same set as #1192 - every caller of `treemacs--execute-button-action` without an explicit `:window`:

- `treemacs-visit-node-no-split`
- `treemacs-visit-node-vertical-split`
- `treemacs-visit-node-horizontal-split`

The split variants benefit specifically because splitting the leftmost window (when treemacs is on the right) is even more disorienting than visiting it - it inserts the new window far from where the user is looking.

## Tests

Added three buttercup specs under `treemacs--window-next-to-treemacs` in `test/treemacs-test.el`:

- adjacent pick with treemacs on the right of multiple normal windows
- adjacent pick with treemacs on the left of multiple normal windows
- fallback to the cycle-and-skip helper when the adjacent slot is dedicated

The full suite (`make test`) is green at 260 / 260.

[docstring]: src/elisp/treemacs-interface.el


rain-64:
Closing for now — this PR's fallback path reuses the cycle-and-skip helper from #1192, which only cycles via `next-window`. With treemacs on the right of multiple non-side windows that's the wrong direction: if `window-in-direction` returns nothing usable, the helper wraps to the leftmost window rather than walking backwards to treemacs's nearest left neighbor. Will rework the helper to cycle in the direction that matches `treemacs-position` and reopen.

